PR #19, opened 2021-06-22
Change "Class" to "Topic"

### LINKED ISSUE #2: Change 'classX' to 'TopicX'.
Title. 
I hate the use of classX since they are not classes, but topics.

jdelamer:
Following closes #2


jameshughes89:
Perhaps also rename the files to `topicX.rst` too. 

This would require some extra searching for links to anything with `classX.rst`, but these could be found with a `grep`

jdelamer:
@jameshughes89 I have change the filenames and the link in the labs, etc.

Issue #21, opened 2021-07-23
Include Lex Fridman googling video


jameshughes89:
If they don't take me seriously, then maybe they'll take him seriously. 

https://www.youtube.com/watch?v=e9C_cgL2150

PR #20, opened 2021-07-06
Update headings

jdelamer:
#1  Update headings following: https://thomas-cokelaer.info/tutorials/sphinx/rest_syntax.html#id8.
As each page was starting with a "section," I updated the subsection consequently.
I also checked that the underlines were of the size of the titles.
All the files are now following the organization below:

```
Section
=======

Subsection
----------
```

Issue #9, opened 2021-06-13
Remove online section details


jameshughes89:
Cut the section 12 (online) details from index and outline.

jameshughes89:
There's also something in class/topic 1 about the two sections.

Issue #1, opened 2021-06-11
Update headings such that they follow the correct heirarchy


jameshughes89:
Currently the files have no consistency with respect to heading levels. Go through and fix them. https://thomas-cokelaer.info/tutorials/sphinx/rest_syntax.html#headings

PR #25, opened 2021-07-23
:mortar_board: Topic1 Review and update

jameshughes89:
### Related Issues or PRs
Related to #20 #21 #9 #1

### What
Do an edit pass over topic 1. A few small changes are made. 

### Why
Haven't done this in a while. 

### Additional Notes
This PR should not suggest Topic1 is now ready.


source/topic1.rst
```diff
@@ -104,20 +106,21 @@ We've already done this once already, but let's try it again:
...
-and you're keen to get the most out of the class, it's worth considering.   
+* Just to be crystal clear: you **do not** need to bring a laptop to class
+* If you have one, and you're keen to get the most out of the class, it's worth considering
    
    
 Is this class easy?
-^^^^^^^^^^^^^^^^^^^
+===================
+
+The quick answer is, no --- it's pretty hard
```
twentylemon, line 123: The quick answer is, it depends. Ease is relative.

source/topic1.rst
```diff
@@ -128,73 +131,78 @@ Is this class easy?
...
+* Unfortunately, intro to CS classes do often have an unusually high fail rate.
 
 
 So I should be terrified?
-^^^^^^^^^^^^^^^^^^^^^^^^^
+=========================
 
-No. Really really no. It's actually not that bad, but you should be prepared to be challenged (but that's why we're all here, right?). 
+* No. 
+* It's actually not that bad, but you should be prepared to be challenged
```
twentylemon, line 146: I'm not sure I would word it like that. Sounds like dunking on those that do end up failing. Perhaps adding something like this is sufficient to set the tone:
> It's actually not that bad _if you put in the effort_

PR #50, opened 2021-09-15
:mortar_board: Topic3 General cleanup edits

jameshughes89:
### What
General clean up here and there. Nothing particularly noteworthy other than cutting useless crap. Most exciting thing is putting math-ie math back in. 

### Why
Bad 'jokes' are bad

### Additional Notes
Will need to double check that the math still renders as I think there _may_ be a change needed in the `conf.py` file.

### DIFF
--- a/source/topic3.rst
+++ b/source/topic3.rst
@@ -301,10 +301,7 @@ Function values
 
     Write a function ``compmag(r,m)`` to compute, and return, the magnitude of a complex number. It should take the real component of the number as parameter ``r`` and the imaginary component as ``m``.
    
-    .. Remember that :math:`|r + mi| = \sqrt{r-2 + m-2}`. Say, does Python have a square root function?
-    .. How would you find it?
-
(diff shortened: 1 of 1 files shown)


jameshughes89:
Just checked. Looks like the `conf.py` file is set correctly.

twentylemon:
> Will need to double check that the math still renders as I think there _may_ be a change needed in the `conf.py` file.

That's why the [actions](https://github.com/jameshughes89/cs101/pull/50/checks) create the website as an actifact.

source/topic3.rst
```diff
@@ -341,8 +334,8 @@ Composition
 
     Figure out the value of ``no_stuff(no_stuff(2,2), (no_stuff(2,2) + no_stuff(4,4)) )`` using only *pen and paper*. No computers!
 	
-Variable scope (not the mouthwash)
-==================================
+Variable scope
+==============
```
twentylemon, line 338: Bad jokes do be being bad, though.

twentylemon (approved):

PR #56, opened 2021-09-17
:mortar_board::lipstick: Topic3 Cleaning up whitespace in statements/expressions

jameshughes89:
### What
Small formatting changes to better match conventions. 

### Why
It's better this way.


twentylemon (approved):
`grep -r ',[^ ]'` will find you these. If you have globstar, you can do `grep -r ',[^ ]' **/*.rst` to only look in rst files.

PR #61, opened 2021-09-22
:mortar_board: Topic4 Clean up end sections wording

jameshughes89:
### What
Cleaned up some wording in the discussions on (a) nesting, (b) libraries, and (c) NumPy. 

### Why
The wording was either wrong, or wrong-ish, or arguable, or nuanced, or needs far more explaining, or bleh

### Where
Only the latter portion of the topic. There may be more cleanup needed for the earlier sections. It was just the latter sections that stood out particularly bad to me. 

### Additional Notes
Most of this, if not all, was vestigial content from back when the content was for UWO's COMP 2120. The focus of 2120 was a little different then a CS 101 course.

### DIFF
--- a/source/topic4.rst
+++ b/source/topic4.rst
@@ -241,8 +241,7 @@ Libraries
 * Most of you are here because you are pragmatic people who want to *get stuff done*
 * The fastest way to *get stuff done* is by leveraging stuff that other people have done.
 * Remember functions? Wouldn't it be awesome if there were huge collections of functions that already existed... and did a lot of the stuff you want to do? 
-* One of reasons we're using Python is because it has a *huge* variety of existing **libraries**/**packages**.
-* No matter what you want to do, there's probably a library that can help you.
+* Python has a *huge* variety of existing **libraries**/**packages**.
 	   
 	   
 NumPy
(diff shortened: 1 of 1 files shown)


source/topic4.rst
```diff
@@ -199,7 +199,7 @@ Nested conditionals
         else:
             print('Third Quadrant')
 			
-* In the above example, we *could* have done it without nesting and using just ``and`` and whatnot, but the nesting does make things easy to follow. 
+* In the above example, we *could* have done it without nesting by using ``and`` and whatnot, but for fun we can nest it
 
 * Again, no limit to how deep you nest... but mind the readability of your code!
```
twentylemon, line 204: Python does actually have a limit to the amount of nesting, it's 20. [StackOverflow question](https://stackoverflow.com/questions/44972719/why-does-python-have-a-limit-on-the-number-of-static-blocks-that-can-be-nested) has some context.

twentylemon (approved):

source/topic4.rst
jameshughes89: Oh shit, I totally remember learning this earlier this year.

jdelamer: Really? Good to know.

jameshughes89: Neat, eh!?

PR #62, opened 2021-09-23
:bomb: Labs Remove 'online section' instructions

jameshughes89:
### What
Remove the details at the end of each lab explaining that the online section must submit their lab via Moodle. 

### Why
There is no online section this year, and these instructions just confuse the students. 

### Where
End of all lab files. It was a WARNING.

### Additional Notes
I was gonna' just comment it out just in case it came back in the future, but (a) I don't think online sections will be back for a while, (b) it's easy to just add the warning back in, and (c) I was too lazy if commenting out the warning would actually work since  the `..` for comments is similar to `.. warning::` for warning, so, idk. Easier to just delete.


twentylemon:
> I was gonna' just comment it out just in case it came back in the future

This is part of what version control is for. Don't leave crap in your code base if it's in history.

twentylemon (approved):

PR #63, opened 2021-09-23
:memo: Add link to course to readme

jameshughes89:
### What
Add a link to csci161 to the readme.

### Why
So people can see it if they want. 

### Where
Top of readme. 

### How
Literally copy/pasted from CS102 and changed a number.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 # cs101
 CS 101: Introduction to Computer Science --- Python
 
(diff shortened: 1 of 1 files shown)


twentylemon (changes requested):
Isn't it cs161?

jameshughes89:
>> Isn't it cs161?

> Literally copy/pasted from CS102 and changed a number.

F

twentylemon (approved):

PR #67, opened 2021-09-28
:bug: Asn1 Fix function header and remove output cells from ipynb

jameshughes89:
### What
Change `:return X:` -> `:param X:` for the `num_starbucks` function header. Also remove the bad executed code from the notebook. 

### Why
Currently in the asn1.ipynb file the function header is wrong and says return when it should say param. There is also some output from cells that were run that are saved in the notebook. These may be confusing. 

### Additional Notes
Oddly, the asn.py file is just fine. :shrug:


twentylemon (approved):
No idea why the diff is 400 pages, but `return` -> `param` checks out.

PR #73, opened 2021-10-01
:bomb: Asn1 Remove code that should not be there

jameshughes89:
### What
Remove excess broken code that were left in cells. Also, for some reason one of the functions they were to write was 3/4ths already done for some reason. 

### Why
Broken

### Additional Notes
I will not merge this until at least Tuesday since the removal of the working part of the one function would be perhaps not very fair to those that have yet to start.


twentylemon (approved):
These `ipynb` files are hard to review.

PR #77, opened 2021-10-13
:bug: Assignment3 Fix link to aliasing

jameshughes89:
### What
Correct the link to aliasing in topic8. 

### Why
It was linking to the old location/website. 

### Where
At the beginning, where I give them a hint (that no one will read) about refs and values and whatnot. 

### How
Replace the link with the correct one (http://modsurski.com/csci161/topic8.html#aliasing)

### Additional Notes
I wish I could link with the `:doc:`, but I can't use the `#` to specify a spot with that. :shrug:

### DIFF
--- a/source/asn3.rst
+++ b/source/asn3.rst
@@ -14,12 +14,11 @@ Assignment #3: I'm only taking CS to make sick video games
    * 2D lists
    * Generalizing rules 
 
+
 .. warning::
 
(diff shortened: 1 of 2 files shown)


twentylemon (changes requested):
> I wish I could link with the `:doc:`, but I can't use the `#` to specify a spot with that. :shrug:

https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#role-ref

PR #79, opened 2021-10-13
Assignment3 Add due date

jameshughes89:
### What
Add a due date for A3. 

### Additional Notes
I would have preferred Nov 8th in 2021, but that's fall reading week, so, no dice.


twentylemon (approved):
Fall reading week? I feel like I knew that was a thing but it still surprises me every time I hear it.

PR #83, opened 2021-10-15
:lipstick: Rename params and add space

jameshughes89:
### What
- Add a space between the parameters.
- Rename `char` -> `character` and `string` to `phrase` params for the functions.

### Why
- Spacing for consistency. 
- Renaming the params since `char` and `string` are keywords in other languages, so this just makes things more different. 

### Where
Lab 6, questions 1 -- 4. 

### Additional Notes
`char` and `string` are fine in python, but this helps make things more distinct. Further, sometimes new students get mixed up with the variable/param names and thinking that, because it's a string, it **has** to be called string, which is not correct.

### DIFF
--- a/source/lab6.rst
+++ b/source/lab6.rst
@@ -9,20 +9,20 @@ We're going to do a bunch of linear searchers to start.
 
 **1**
 
-Write a function ``char_is_in_while(char,string)`` that returns ``True`` if the character char appears in the string ``string``, otherwise return ``False``. This function **must** use a *while* loop. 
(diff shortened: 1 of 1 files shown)


jameshughes89:
> > `char` and `string` are fine in python, but this helps make things more distinct. Further, sometimes new students get mixed up with the variable/param names and thinking that, because it's a string, it **has** to be called string, which is not correct.
> 
> Eh. I don't really agree with the reasoning, arguments still have types, and functions inherently require a specific type. Naming is one way to inform callers of that type. But whatever.

Totally agree --- that part of the change is not necessary, but based on what I see in the trenches (lab), it does mess a few of the students that perhaps are a tad behind. 

That said, I totally think I will now do type hints. Idk where to start, but soon-ish would be awesome.

PR #84, opened 2021-10-15
:lipstick::bug: Lab6 Fix formatting --- add space

jameshughes89:
### Related Issues or PRs
Broke things in #83 

### What
Add a space between code like back quote area

ALSO 

:new: Change `char` to `character` in function defs since I missed it. 

### Why
Missed space made it format wrong. 

### Where
Q2, between character and appears.


jameshughes89:
> My effort has been used for an EXAMPLE!?? angry

it was used for love :heart:

PR #109, opened 2022-03-08
:robot: :memo: Restructure Project

jameshughes89:
### What
Restructure the file structure for the course content

### Why
Cleaner this way, but will also make it easier to "scale" (for lack of a better term). 

### Where
Everywhere 

### How
Pain

### Testing
Yes.


site/index.rst
```diff
@@ -142,75 +142,78 @@ Getting Started
...
-    topic12
-    topic13
-    topic13b
-    topic14
-    topic15
-    topic16  
-    topic17
-    topic18
-    topic19
+    topics/intro/intro
```
twentylemon, line 154: Minor, does rst support "index" files? Like when you navigate to `whatever.com`, you actually go to `whatever.com/index.html`. Would be a nice cleanup here to only say `topics/intro`.

jameshughes89: I had tried this. Sadly, no dice.

jameshughes89: actually --- hold on, I have one last idea.

jameshughes89: the new idea failed too

PR #112, opened 2022-03-09
:new: Start Function Topic

jameshughes89:
### What
Write some functions based on the functions topic. 

### Why
Gotta' start somewhere, and this is the first topic with anything worth having. 

### Testing
Yes

### Additional Notes
I'll look into having github run tests auto somehow. 
This is not complete. I just wanted to get something down. 

`tester` is the current way I am thinking of them doing tests. Since they will be using colab, pytest isn't too workable.


twentylemon:
> `tester` is the current way I am thinking of them doing tests. Since they will be using colab, pytest isn't too workable.

python has [unittest](https://docs.python.org/3/library/unittest.html) built in. pytest is _probably_ better since you don't need to know classes and assertions are a lot cleaner by using `assert` instead of an inherited assertion method.

.gitignore
```diff
@@ -1,6 +1,9 @@
 # All built stuff
 /out/
 
+# Cache stuff
+__pycache__
+
```
twentylemon, line 6: https://github.com/github/gitignore/blob/main/Python.gitignore

I usually opt for `curl https://raw.githubusercontent.com/github/gitignore/main/Python.gitignore >> .gitignore` -- you can get the raw link from the github page. There's usually some editing to do afterwards as well, they have comments explaining when you may want to ignore particular things.

src/functions_topic.py
```diff
@@ -0,0 +1,45 @@
...
+    @param string2: Second string of the concatenation
+    @return: The concatenation of string1 and string2
+    """
+    concatenated = string1 + string2
+    return concatenated
+
+
+# Tests for the concatenate_strings method
+tester.test_function("", concatenate_strings, "", "")
+tester.test_function("ab", concatenate_strings, "a", "b")
```
twentylemon, line 21: This is... not normal. Using a higher order function to test is pretty rough for potentially people with literally no coding experience.

test/functions_topic_test.py
```diff
@@ -0,0 +1,21 @@
...
+from src.functions_topic import *
+
+
+@pytest.mark.parametrize("string1, string2, expected",
+                         [("", "", ""),
+                          (" ", " ", "  "),
+                          ("a", "b", "ab"),
+                          ("Hello", "World", "HelloWorld")])
+def test_concatenate_strings_returns_concatenated_string(string1, string2, expected):
+    assert concatenate_strings(string1, string2) == expected
```
twentylemon, line 11: Keep it simple. `parametrize` makes sense for actual python projects to reduce bloat, but it's really only useful if you know what you're doing in the first place. For learning, I'd much rather see completely braindead tests.
```python
def test_concat_a_b():
    assert concat("a", "b") == "ab"

def test_concat_empty():
    assert concat("", "") == ""
````

Comment (jameshughes89) [reply test/functions_topic_test.py:11]: In case it's not clear, the idea is
1. The hacked test function I made is for them learning the idea with unit testing but with something that would work on Colab.
2. The pytest is just for the repo

That said, it's a good point about the tests being simple AF anyways for learning purposes _if_ they go digging on the repo.

File: src/functions_topic.py (hunk 2 above)
Comment (jameshughes89) [reply src/functions_topic.py:21]: Any thoughts on an alternative? I figured this would be easiest for them, but there is defo a lot of "magic code" in here.

Comment (jameshughes89) [reply src/functions_topic.py:21]: I suppose just writing out the assert on their own would work. 

`assert "ab" == vincat("a", "b")`

Comment (twentylemon) [reply src/functions_topic.py:21]: `assert`s don't do anything in normal execution typically -- I'm not sure what python's behaviour is there, but I imagine it'd be the same as many other languages.

File: test/functions_topic_test.py (hunk 3 above)
Comment (twentylemon) [reply test/functions_topic_test.py:11]: > The pytest is just for the repo

bruh onboard black, your python code is ugly as sin

File: src/functions_topic.py (hunk 2 above)
Comment (jameshughes89) [reply src/functions_topic.py:21]: I think it throws an assertion error, but I'll look into this when I'm supposed to be lecturing.

Comment (twentylemon) [reply src/functions_topic.py:21]: It likely only does so when debugging flags are enabled.

File: test/functions_topic_test.py (hunk 3 above)
Comment (jameshughes89) [reply test/functions_topic_test.py:11]: Wow, rude. 

But yeah. I'll get that up and running with a PR doing `setuptools` et al.

File: src/functions_topic.py (hunk 2 above)
Comment (twentylemon) [reply src/functions_topic.py:21]: > Any thoughts on an alternative?

Do you touch on classes and inheritance? Maybe testing can be introduced after that topic and you use the built-in `unittest`.

Comment (jameshughes89) [reply src/functions_topic.py:21]: Inheritance isn't covered, and testing is one of the last topics. 

Additionally, it appears to be a nightmare getting `unittest` to work on colab and requires mounting your Google drive (that said, a quick googz seems to suggest otherwise --- maybe this was changed recently-ish?). I'll do a little more digging here. 

Sadly, still, it would have to be hit late in the course and it would be good to cover testing earlier.

Comment (jameshughes89) [reply src/functions_topic.py:21]: Looks like colab will throw the error regardless. 

image

Comment (twentylemon) [reply src/functions_topic.py:21]: Nice. Either they run in debug mode or python just always behaves that way. My guess is the former, without looking into it at all.

Comment (jameshughes89) [reply src/functions_topic.py]: [Looks like `unittest` is actually easy in colab.](https://colab.research.google.com/github/damorimRG/practical_testing_book/blob/master/testgranularity/unittesting.ipynb)

That said, it requires classes which is a late course topic. Solution --- Use simple asserts to start. When objects are hit, we can use `unittest`.

Comment (twentylemon) [reply src/functions_topic.py]: That's probably reasonable enough. Think it makes sense just to get the idea across of semi automated testing, then be like "yo dawg this is actually your life" much later.

File: src/functions_topic.py [hunk 4]
```diff
@@ -0,0 +1,42 @@
...
+    @param string2: Second string of the concatenation
+    @return: The concatenation of string1 and string2
+    """
+    concatenated = string1 + string2
+    return concatenated
+
+
+# Tests for the concatenate_strings method
+assert "" == concatenate_strings("", "")
+assert "ab" == concatenate_strings("a", "b")
```
twentylemon, line 18: Just a note that only the first failure will be shown with tests like this. Thats probably fine for learning purposes, but definitely a limitation to be aware of.

twentylemon (approved):

src/functions_topic.py
jameshughes89: Good point. 🤔

jameshughes89:
> > `tester` is the current way I am thinking of them doing tests. Since they will be using colab, pytest isn't too workable.
> 
> python has [unittest](https://docs.python.org/3/library/unittest.html) built in. pytest is _probably_ better since you don't need to know classes and assertions are a lot cleaner by using `assert` instead of an inherited assertion method.

I've opted to switch to `unittest` since:
1. The simple asserts will be used as their simple tests
2. It will be used later
3. It doesn't require anything extra
4. It's not THAT bad --- sure there will be "magic code", but it's minimal-ish at the level it will be used.

test/functions_topic_test.py
```diff
@@ -0,0 +1,17 @@
+import unittest
+from src.functions_topic import *
```
twentylemon, line 2: onboard `isort` as well, when you get around to that. And maybe `flake8`. Just take duckbot's config.

It's usually bad practice to do a `*` import, flake8 would yell at you for that.

twentylemon (approved):

PR #113, opened 2022-03-10
🤖 Workflow Add unit tests for sample code to workflow

jameshughes89:
### What
Add a job to the workflow for unittesting the sample code. 

### Why
It would be nice to check for correctness. 

### Testing
I'll see what happens on this PR 😬


.github/workflows/sphinx-build.yml
```diff
@@ -11,7 +10,25 @@ on:
...
+      run: |
+        python -m pip install --upgrade pip setuptools wheel
+        python -m pip install -r requirements.txt
+    - name: unittest sample code 
+      run: |
+        python -m unittest discover -s .\test\ -p "*_test.py"
+
   sphinx:
+    needs:
+    - test
```
twentylemon, line 31: Sphinx build doesn't really need unit tests to pass. I'd move this to be a dependency of `deploy`, which will have the added benefit of being executed in parallel then.

jameshughes89:
image
nice

twentylemon (approved):

PR #115, opened 2022-03-10
:robot: :memo::telescope::lipstick Use setuptools and isort flake8 black

jameshughes89:
### What
1. Make use of isort black flake8
2. use setup tools and remove requirements
3. run formatter
4. update readme
5. update workflows

### Why
Nice clean code. Also better workflows. 

### How
Basically copied duckbot. 

### Testing
~Kinda'~
yes

### Additional Notes
~Im gonna make changes to the workflows in draft version to make sure things work.~

### DIFF
--- a/.github/workflows/sphinx-build.yml
+++ b/.github/workflows/sphinx-build.yml
@@ -42,11 +42,20 @@ jobs:
     - name: Set up Python
       uses: actions/setup-python@v2
       with:
-        python-version: 3.9
-    - name: Install Dependencies
+        python-version: 3.8
+    - name: Setup Virtual Environment Cache
+      id: venv
+      uses: actions/cache@v2
+      with:
+        path: venv
+        key: ${{ env.pythonLocation }}-venv-${{ hashFiles('setup.py', 'pyproject.toml') }}
+    - name: Install Packages
+      if: steps.venv.outputs.cache-hit != 'true'
       run: |
-        python -m pip install --upgrade pip setuptools wheel
-        python -m pip install -r requirements.txt
+        python -m venv --clear venv
+        . ./venv/bin/activate
+        pip install --upgrade pip setuptools wheel
+        pip install --editable .
     - name: Sphinx Build
       run: |
         sphinx-build -b html site/ out/
--- a/.github/workflows/sphinx-build.yml
+++ b/.github/workflows/sphinx-build.yml
@@ -10,7 +10,7 @@ on:
     - main
 
 jobs:
-  flake8:
+  format:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
@@ -22,10 +22,9 @@ jobs:
       run: |
         python -m pip install --upgrade pip setuptools wheel
         python -m pip install --editable .
-    - name: flake8 on src and test
+    - name: Format Check
       run: |
-        flake8 ./src/
-        flake8 ./test/
+        format
         
   test:
     runs-on: ubuntu-latest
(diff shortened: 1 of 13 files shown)


jameshughes89:
Spam should be done.

jameshughes89:
flake8 & test combined. Should be done for realz this time.

.github/workflows/sphinx-build.yml
```diff
@@ -1,4 +1,4 @@
-name: Unittest Sphinx Build Deploy
+name: Flake8 Test Sphinx Build Deploy
```
twentylemon, line 1: Flake, black, isort etc all fall into the testing bucket imo. They're verifications made against the code. Name is looking quite odd as is.

setup.py
```diff
@@ -0,0 +1,32 @@
...
+            "flake8-black==0.3.2",
+            "flake8-isort==4.1.1",
+            "isort==5.10.1",
+            "mdformat==0.7.13",
+            "mdformat-gfm==0.3.5",
+            "mdformat-black==0.1.1",
+            "sphinx==4.0.1",
+            "sphinx-rtd-theme==1.0.0",
+        ],
+        entry_points={"console_scripts": [f"format = setup:{run_code_formatters.__name__}"]},
```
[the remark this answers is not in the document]
jameshughes89: I spent a crazy amount of time trying to, but setup tools was pissy about sphinx not existing. Similarly with flake8, which is strange because it worked in the formatter script.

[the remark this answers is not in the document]
twentylemon: That's because sphinx isn't available at build time. The reason the formatters work is you're not actually referencing them, just firing up a shell that will eventually reference them. You could use the same workaround, or add sphinx as a build dependency in `pyproject.toml`, similar to how [duckbot installs nltk data](https://github.com/duck-dynasty/duckbot/blob/main/setup.py#L15-L25).

setup.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+if __name__ == "__main__":
+    setup(
+        name="cs101",
+        version="1.0",
+        url="https://github.com/jameshughes89/cs101",
+        python_requires=">=3.8",
+        packages=find_packages(),
+        install_requires=[
```
twentylemon, line 19: Should probably set up dependabot to help manage these for you as well. You should be able to rip it wholesale off of [duckbot](https://github.com/duck-dynasty/duckbot/blob/main/.github/dependabot.yml).

setup.py
```diff
@@ -25,7 +25,7 @@ def run_code_formatters():
             "mdformat==0.7.13",
             "mdformat-gfm==0.3.5",
             "mdformat-black==0.1.1",
-            "sphinx==4.0.1",
+            "sphinx==4.4.0",
```
jameshughes89, line 28: #dependabot

setup.py
```diff
@@ -28,5 +34,10 @@ def run_code_formatters():
             "sphinx==4.4.0",
             "sphinx-rtd-theme==1.0.0",
         ],
-        entry_points={"console_scripts": [f"format = setup:{run_code_formatters.__name__}"]},
+        entry_points={
+            "console_scripts": [
+                f"format = setup:{run_code_formatters.__name__}",
+                f"validate = setup:{run_code_verification.__name__}",
```
jameshughes89, line 40: validate because verify is reserved on windows

setup.py
jameshughes89: I'll make a new PR for that

PR #128, opened 2022-03-24
:new: 🎓 Move and Enhance Input

jameshughes89:
### What
Move Input to its own topic following the variables topic

### Why
Feels like the variable topic included too much and it seems like a reasonable thing to stand alone


twentylemon (approved):
Please try to keep moving files as their own PR, then the changes as another PR. It's more difficult to review when the whole file is green.

PR #132, opened 2022-03-24
💄 Values Topic Move Print in Title and Fix Whitespace

jameshughes89:
### What
Put print at the start of the topic title. Also add point about using `print` already. 

### Why
It comes before the other bits

### DIFF
--- a/site/topics/values-variables-types/values-variables-types.rst
+++ b/site/topics/values-variables-types/values-variables-types.rst
@@ -1,5 +1,5 @@
 *******************************
-Values, Variables, Types, Print
+Print, Values, Variables, Types
 *******************************
 
 .. admonition:: Activity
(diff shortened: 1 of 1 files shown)


site/topics/values-variables-types/values-variables-types.rst
```diff
@@ -66,8 +66,8 @@ Languages
    
 * The world is a screwed up, scary, place (for mathematicians, anyways). If you want to fall down this particular rabbit hole:
     * https://en.wikipedia.org/wiki/G%C3%B6del%27s_incompleteness_theorems
-	* http://en.wikipedia.org/wiki/Principia_Mathematica
-	* http://en.wikipedia.org/wiki/Computability   
+    * http://en.wikipedia.org/wiki/Principia_Mathematica
+    * http://en.wikipedia.org/wiki/Computability
```
twentylemon, line 70: Hell, since we're saying the universe is scary, it turns out that the universe must be [probabalistic in nature](https://en.wikipedia.org/wiki/Bell%27s_theorem), at least with our current understandings of math.

twentylemon (approved):

site/topics/values-variables-types/values-variables-types.rst
jameshughes89: probably

jameshughes89: I actually already yeeted this in a previous PR --- I guess I just got carried away with fixing spaces/tabs.

twentylemon: Wow.

PR #134, opened 2022-03-25
:lipstick: Functions Rework Part 2

jameshughes89:
### What
Rework function topic more up to and EXCLUDING abstraction.

### DIFF
--- a/site/topics/functions/functions.rst
+++ b/site/topics/functions/functions.rst
@@ -87,9 +87,8 @@ Function Parameters
         return temp_in_fahrenheit
 
 
-* Note carefully the parameter (``a_parameter``) in the definition of ``my_function``
-* When you are defining a function, you want the function to be very *general*
-    * You want it to work with *any possible* parameter that someone might want to give it
+* You have already been specifying the parameter in the ``print`` function
+    * The value you want printed out is the value you are setting for the parameter that ``print`` takes
 
 * In reality, you have been using this idea in math class for years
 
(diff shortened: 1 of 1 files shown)


site/topics/functions/functions.rst
```diff
@@ -67,74 +67,139 @@ Calling a Function
...
+        return d
+
+
+* If I were to call this function with ``square_of_sum(2, 3)``, Python handles the execution like this
+    #. Python will check to see if it knows about a function called ``square_of_sum``
+    #. Python takes the values supplied to it when called (``2`` and ``3``) and assigns them to their respective parameters
+        * ``a = 2`` and ``b = 3``
+    #. The sum of ``a`` and ``b`` is put into a variable ``c``
+    #. The variable ``c`` is multiplied with itself (effectively squaring it) and the result is assigned to ``d``
+    #. The function returns the value associated with ``d``
```
twentylemon, line 131: May be worth mentioning the fact that the program flow jumps from the call site to whatever function is called. If `__main__` calls a function, program flow won't return to `__main__` until that function returns a value. This is briefly touched on later, but it seems to fit here.

twentylemon (approved):

site/topics/functions/functions.rst
jameshughes89: Moving it up is a great idea. I'ma do that.

jameshughes89: I'll move it into the section in a seperate PR though.

PR #137, opened 2022-03-29
💄 Functions Rework Part 3

jameshughes89:
### What
Rework the abstraction section
Remove unnecessary sections from the latter part of the topic 

### Why
Cleaning and, although making libraries and optional params are cool and all, they are not necessary right here.


twentylemon (approved):

jameshughes89:
From @twentylemon  offline

> You could say that abstraction is why programming is hard. Defining new abstractions requires a deep understanding of the problem space

Which I think is a great idea. Point out that this isn't some trivial thing for them to pick up --- this is a helluva skill to learn.

PR #147, opened 2022-04-06
💄 Functions Rework Composition Section

jameshughes89:
### What
Rework the Composition of functions section

### Why
It's better this way.

### DIFF
--- a/site/topics/functions/functions.rst
+++ b/site/topics/functions/functions.rst
@@ -374,26 +374,37 @@ Format of a Function
 Composition
 ===========
 
-* Python functions can be *composed* just like mathematical functions.
-* We've already seen ``print`` composed with ``do_stuff``
-* We can nest functions, too:
-    >>> do_stuff(do_stuff(2, 2), do_stuff(2, 2))
-    72
-* If you get confused tracing nested functions, just remember:
-    * Functions get *evaluted* and turned into values
-    * Find a function you can evaluate
-    * Evaluate it
-    * Cross out the function and replace it with the *value* it returns
-    * Keep doing this until you're down to one value.
+* We can compose functions of other functions
+
+    * Like in the ``criss_cross_concatenation`` function that makes use of ``concatenate_strings``
+
+* Since functions often return a useful value, we can nest function calls in statements
+
+    * ``square_of_sum(2, 2) + square_of_sum(3, 3)``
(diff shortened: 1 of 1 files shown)


site/topics/functions/functions.rst
```diff
@@ -374,26 +374,37 @@ Format of a Function
...
-* We can nest functions, too:
-    >>> do_stuff(do_stuff(2, 2), do_stuff(2, 2))
-    72
-* If you get confused tracing nested functions, just remember:
-    * Functions get *evaluted* and turned into values
-    * Find a function you can evaluate
-    * Evaluate it
-    * Cross out the function and replace it with the *value* it returns
-    * Keep doing this until you're down to one value.
+* We can compose functions of other functions
```
twentylemon, line 377: of -> from ? from seems to fit better

twentylemon (approved):

site/topics/functions/functions.rst
jameshughes89: hmmmmmmmmmm. I don't disagree, but I also don't disagree with the other

jameshughes89: how about

`* Functions can be composed of other functions`

twentylemon: Still _from_ fits a little better, imo.

jameshughes89: Is the following a reasonable comparison: _This cake is composed of flower, water, sugar, salt..._

jameshughes89: or is that _from_ too? I really don't know.

twentylemon: /shrug

maybe students will open a pr

jameshughes89: According to people on the internet, (it's of)[https://english.stackexchange.com/questions/56985/what-should-i-use-composed-of-or-composed-from#:~:text=Composed%20of%20is%20standard%20from,some%20other%20such%20contextual%20situation.&text=Neither%20composed%20of%20nor%20composed%20from%20is%20a%20phrasal%20verb.].

I guess it's what they call a _collocation _ --- we just kinda' like using them together even though from is totally fine really.

twentylemon: I blame society.

Issue #85, opened 2021-10-15
:new: Type Hints


jameshughes89:
### Related Issues or PRs
As mentioned in #83 

### What 
Start using type hints

### Why
1. Helps them think about types and seperating them from the variable names
2. Python's _not-typed-kinda-ness_ seems great to students at first, but in reality, it sucks bad --- type correctness is good. e.g. In the context of kattis since they use `input` which always gives you a string, but they forget and scratch their head forever.  
3. They will need to get used to typed stuff

### Where
I'm not sure exacrlt. I could see an argument for starting basically at or very near the beginning of the course, or somewhere along the way, perhaps Topic 13 & 13b where Objects are brought up.

jameshughes89:
Type hints are explicitly included now.

Issue #76, opened 2021-10-12
:mortar_board: Unit testing


jameshughes89:
### What 
I would love to add examples of unit tests to some of the mid to later topics. Nothing huge, nothing scary, but something to help with getting them thinking about testing their functions. 

### Why
1. It will help getting them thinking about how to test their functions, regardless of if they write the tests or just think about how to test them manually. 
2. It will give me a chance to figure out python unit testing
3. It's a good skill for them to learn
4. It will be good prep for 102, which uses unit testing. 

### Where
Probably around the debugging topic? Or perhaps once we introduce linear search since it's got the classic bug of:

```python
def char_is_in(char, string):
    count = 0
    while count < len(string):
        if string[count] == char:
            return True
        else:                    # bad
            return False         # bad
        count = count + 1
```

If introduced here, perhaps more of a _hey look, this is a thing..._, and then later at debug it could go into actual details. 

### How
Writing some tests. 

### Additional Notes
Unlikely to make it for Fall 2021, but going forward, this would be good.

jameshughes89:
[Here is where the text introduces it](http://openbookproject.net/thinkcs/python/english3e/fruitful_functions.html#unit-testing) --- after functions with returns & conditionals. In other words, it can be thrown in a little earlier. 

They do it without any system --- they use the following:

```python
import sys

def test(did_pass):
    linenum = sys._getframe(1).f_lineno   # Get the caller's line number.
    if did_pass:
        message = "Test at line {0} ok.".format(linenum)
    else:
        message = ("Test at line {0} FAILED.".format(linenum))
    print(msg)

def test_suite():
    test(expected_A == some_function(A))
    test(expected_B == some_function(B))
    test(expected_C == some_function(C))
    test(expected_D == some_function(D))
    test(expected_E == some_function(E))
```

This is more or less taken from the textbook, but I'm liking this system, or something similar to it, because:
1. Something about unit testing and forrest through trees
2. Limited functionality, but good enough
3. Gets them thinking about testing early
4. Can do on Google colab easy

If I do this, I'll probably switch out line number for some message --- `test(did_pass, message)`

jameshughes89:
As @twentylemon pointed out in #112, this way of doing tests may be missing the mark. If the idea is to make it simpler for them, this is kinda making it not that. 

Solution:
1. At this point `assert` can be used for simple tests
[...]

PR #149, opened 2022-04-06
:new: 🎓 Create New Topic for Testing and Type Hints

jameshughes89:
### Related Issues or PRs
From #76 and #85

### What
Create a new empty topic page for testing and type hints.

### Why
1. I like the idea of them being more explicitly aware of types earlier since it helps reduce errors
2. Testing is big. I want them to be more used to it. It will help them. I can make use of "tests" in their assignments too. They will be more prepped for the following course. They also said in 2nd semester that they wish they were aware of unit testing earlier. 

### Additional Notes
Leaving issues open since they should be carried throughout the remainder of the content.

### DIFF
--- a/site/topics/testing/testing.rst
+++ b/site/topics/testing/testing.rst
@@ -1,6 +1,6 @@
-*****************
-Testing and Hints
-*****************
+********************************
+Testing Your Code and Type Hints
+********************************
 
 	
 For Next Class
--- a/site/topics/testing/testing.rst
+++ b/site/topics/testing/testing.rst
@@ -1,0 +1,9 @@
+*****************
+Testing and Hints
+*****************
+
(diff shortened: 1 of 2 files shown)


site/topics/testing/testing.rst
```diff
@@ -0,0 +1,9 @@
+*****************
+Testing and Hints
+*****************
```
twentylemon, line 3: Maybe _Testing Your Code and Type Hints_

Title sounds more like hints for the upcoming exam.

twentylemon (approved):

site/topics/testing/testing.rst
jameshughes89: hahaha. Good point.

PR #150, opened 2022-04-07
🆕 Functions Topic Code With Tests

jameshughes89:
### What
Add the code from the topic to a .py file with simple tests. 
Add proper tests for the function topic code

### Testing
Yes

### Additional Notes
Format was run
Turns out the `unittest` assertions produce fine messages, so making custom messages was perhaps overkill. 🤷

### DIFF
--- a/src/functions_topic.py
+++ b/src/functions_topic.py
@@ -2,9 +2,9 @@ def celsius_to_fahrenheit(temp_in_celsius: float) -> float:
     """
     Convert a temperature from Celsius units to Fahrenheit units.
 
-    @rtype: float
-    @param temp_in_celsius: The temperature in Celsius to be converted.
-    @return: The temperature in Fahrenheit.
+    :rtype: float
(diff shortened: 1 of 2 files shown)


twentylemon:
> Turns out the `unittest` assertions produce fine messages, so making custom messages was perhaps overkill.

I'd avoid custom messages at outset. It adds unnecessary overhead for the time being. I'd stick with showing how to write and run tests first, then later include error messages just to show that it's possible and maybe useful.

jameshughes89:
> > Turns out the `unittest` assertions produce fine messages, so making custom messages was perhaps overkill.
> 
> I'd avoid custom messages at outset. It adds unnecessary overhead for the time being. I'd stick with showing how to write and run tests first, then later include error messages just to show that it's possible and maybe useful.

I feel ya.

jameshughes89:
yeeted

twentylemon (approved):

PR #153, opened 2022-04-07
🎓 Testing Testing With Assertions

jameshughes89:
### Related Issues or PRs
#76 

### What
Describe testing with assertions and what happens when they fail. 

### How
To show what happens when it fails, I wrote a broken function.

### DIFF
--- a/site/topics/testing/testing.rst
+++ b/site/topics/testing/testing.rst
@@ -39,6 +39,19 @@ Writing Tests
 
 * Checking zero is perhaps not necessary, but it is somewhat of a peculiar case since zero doesn't have a sign
 
+* To do these tests, we could just call the function and check the output ourselves
+* But we want to automate this process a little
+* Instead, we will make use of assertions
+
+.. code-block:: python
+    :linenos:
+
+    assert 5 == abs(5)
+    assert 5 == abs(-5)
+    assert 0 == abs(0)
+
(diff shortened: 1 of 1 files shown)


site/topics/testing/testing.rst
```diff
@@ -25,6 +25,55 @@ Testing
 Writing Tests
 -------------
 
+* For simplicity, we will keep our testing strategy to assertions about the data
+* Let's say we want to test the absolute value function ``abs``
+
+    * This function is provided to you by Python, so there is no actual need to test it here
+
+* If we are to consider how we could test the ``abs`` function, we want to check that the function output (returns) matches what it should be based on the input (parameters)
```
twentylemon, line 33: So common thing people get wrong here, _parameters_ are part of the function's interface, _arguments_ are the input to a function invocation. `f(x)`, `x` is the parameter, `f(2)`, `2` is the argument. You're doing case testing here, so it should probably be arguments here. There exists better strategies for testing where you actually test the interface of a function, but they're well beyond the scope of this course.

twentylemon (approved):

site/topics/testing/testing.rst
jameshughes89: Legit didn't know there was an actual difference.

jdelamer: Wow, we never stop learning new things.

PR #154, opened 2022-04-08
🐛 Functions Comments emphasize `:`

jameshughes89:
### What
Fix the docstring details --- change the first `:` to an `@`.

### Why
~I guess the `@` is more correct? At least that's what pycharm says, so I trust that.~
PyCharm seems to suggest the `@`, but I can't find details on that actually being standard or anything. Either way, I'll go with what PyCharm says as it likely has more method to the madness than I do.

### DIFF
--- a/site/topics/functions/functions.rst
+++ b/site/topics/functions/functions.rst
@@ -529,8 +529,8 @@ Docstring
         """
         Convert a temperature from Celsius units to Fahrenheit units.
 
-        @param temp_in_celsius: The temperature in Celsius to be converted.
-        @return: The temperature in Fahrenheit.
+        :param temp_in_celsius: The temperature in Celsius to be converted.
+        :return: The temperature in Fahrenheit.
         """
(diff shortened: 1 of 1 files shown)


jdelamer:
Apparently, we should use `:` isntead of `@`.
It is recommended by the pep-287: https://peps.python.org/pep-0287/

twentylemon:
Pycharm defaults to [epytext](http://epydoc.sourceforge.net/epytextintro.html) format, not rst. It's [configurable](https://www.jetbrains.com/help/pycharm/documenting-source-code.html#python-docs) though. Given the notes are all rst, probably makes the most sense to stick with the rst conventions.

jameshughes89:
Good catch my dudes. 

Changed back.

jameshughes89:
Note, I added points emphasizing the `:`, so the PR isn't useless.

twentylemon (approved):

PR #155, opened 2022-04-09
📝 Editor Config File

jameshughes89:
### What
Add editor config file. `format` will still be used, but at least this hedges.


twentylemon (approved):
k

PR #156, opened 2022-04-09
🎓 Car Rental Topic Code

jameshughes89:
### What
Put the code from the topic into a script

### Testing
Yes

### Additional Notes
1. It may still change once I get to the topic content, but this at least starts it out consistent. 
2. Originally I had it set up so `average_kms_per_day` called `total_kms` so it would be a total of 3 functions deep. I ditched this as (a) I don't think it's necessary to explicitly show, and (b) it may end up being just a little confusing.

### DIFF
--- a/src/car_rental_topic.py
+++ b/src/car_rental_topic.py
@@ -1,4 +1,4 @@
-def total_kms(odometer_start, odometer_finish):
+def total_kms(odometer_start: float, odometer_finish: float) -> float:
     """
     This function calculates the total number of kilometers driven based
     on starting and ending odometer readings.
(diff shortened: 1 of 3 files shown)


src/car_rental_topic.py
```diff
@@ -0,0 +1,113 @@
...
+    @return: The total kms driven
+    """
+
+    return odometer_finish - odometer_start
+
+
+assert 0 == total_kms(0, 0)
+assert 100 == total_kms(0, 100)
+assert -100 == total_kms(100, 0)
+assert 100.5 == total_kms(100.5, 201)
```
twentylemon, line 18: Why include assertions and `unittest` tests?

twentylemon (approved):

src/car_rental_topic.py
jameshughes89: - Assertions are the ones they will see at this stage. 
- Unit tests are the "real" tests that also serve as reference for the keeners that want to see the REAL way
- Later in the course, the unittests will be shown as it does work on COLAB.

twentylemon: Got it. Makes sense

PR #158, opened 2022-04-11
🎓 Testing Square Sum Example

jameshughes89:
### What
Show some tests for `square_of_sum` and discuss them. Also took the opportunity to briefly talk about floating point precision. 

### Why
Seeing examples of the tests may be one of the better ways to learn them. 

### Testing
Eff that

### DIFF
--- a/site/topics/testing/testing.rst
+++ b/site/topics/testing/testing.rst
@@ -78,6 +78,32 @@ Writing Tests
 Square of Sums Example Tests
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
+ .. code-block:: python
+    :linenos:
+
+    def square_of_sum(a, b):
+        """
+        Calculate the square of the sum of the two provided numbers.
+        E.g.
+            square_if_sum(2, 3) -> 25
+
+        :param a: First number
+        :param b: Second number
+        :return: The square of the sum of a and b
+        """
+        c = a + b
+        d = c * c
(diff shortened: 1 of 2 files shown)


site/topics/testing/testing.rst
```diff
@@ -78,6 +78,89 @@ Writing Tests
...
+    The simple way to address this issue is to check that the *absolute difference* between the expected answer and
+    the function's result is less than some threshold. In the above example, the absolute difference between the
+    correct answer and what Python says is ``0.000000000000003``, which is a tiny difference; the numbers are nearly
+    identical. I chose a threshold of ``0.001``, so if the absolute difference between the expected and actual result is
+    less than that threshold, I will consider that a passed test.
+
+    The choice of the threshold will depend on the situation. Above I could have picked a much smaller number and the
+    test would have passed. But imagine you are a chemist using instruments that can measure to the nearest milliliter.
+    There would be no sense testing beyond a difference of ``0.001`` liters since you cannot get beyond that precision
+    in real life with those instruments.
```
twentylemon, line 145: This exists in the floating point representation, it's called epsilon. In python, you get it from `sys.float_info.epsilon`. This is the typical offset value you use when testing float point equality.  
Probably not worth mentioning at this point, kinda detracts from the testing point. Maybe worth a bullet point later in the course.

twentylemon (approved):

PR #159, opened 2022-04-11
🎓 Testing C to F Tests and Automation

jameshughes89:
### What
- Include another example of testing (`celsius_to_fahrenheit`)
- Mention `unittest`, but that's it. 

### Why
The more the merrier.

### DIFF
--- a/site/topics/testing/testing.rst
+++ b/site/topics/testing/testing.rst
@@ -184,6 +184,7 @@ Celsius to Fahrenheit Example Tests
     # Tests for celsius_to_fahrenheit function
     assert 32 == celsius_to_fahrenheit(0)
     assert -40 == celsius_to_fahrenheit(-40)
+    assert 23 == celsius_to_fahrenheit(-5)
     assert 86 == celsius_to_fahrenheit(30)
     assert 89.6 == celsius_to_fahrenheit(32)
     # To address precision issues, we can look for a sufficiently small difference between the expected and actual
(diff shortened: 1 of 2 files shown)


site/topics/testing/testing.rst
```diff
@@ -165,12 +165,57 @@ Square of Sums Example Tests
...
+
+
+* Above is a series of assertion tests for the ``celsius_to_fahrenheit`` function
+* Notice the key, broad tests
+
+    * Input of zero
+    * Negative input
+    * Positive input
+    * Input is a float
+    * Output is a float
```
twentylemon, line 200: I'd argue `-40` is a special case and does not cover the general negative number case. It's the only point where the two scales are equal, which sounds like an edge case to me.

_Test driven development_ says you actually write some failing tests first, before writing the code that makes the tests pass. Once the tests all pass, either you add more failing tests and iterate to get the full behaviour you want, or you're done. I'm not one to say anyone should follow TDD (I don't enjoy inflicting pain on others), but there is a benefit to at least thinking about tests _first_. Thinking about the problem you're solving before there's a solution, I'll just assert, is very helpful.

Not really applicable here, but I also say that tests are the first clients of your code. If you find something hard to test, it's likely hard to use.

jameshughes89: > Not really applicable here, but I also say that tests are the first clients of your code. If you find something hard to test, it's likely hard to use.

image

twentylemon (approved):

twentylemon (approved):

PR #163, opened 2022-04-11
🎓 Type Hints Part 2

jameshughes89:
### What
Show what type hints do and don't do

### Why
They're great and all, but still don't actually do much.

### Additional Notes
I was going to have a subsection talking about static typing in other languages, but figured that would just be an anecdote while lecturing or something.


site/topics/testing/testing.rst
```diff
@@ -242,10 +242,41 @@ Setting Type Hints
...
+      :return: The sum of the two numbers
+      """
+      return a + b
+
+    x = input("First number: ")
+    y = input("Second number: ")
+    result = add_together(x, y)
+    print(result)
+
+* In the above example, despite type hints being included, it would behave the same way as it did without type hints
```
twentylemon, line 279: Maybe worth being explicit here, saying it'd concatenate the two strings.

twentylemon (approved):

PR #165, opened 2022-04-21
🎓 Car Rental Rework Intro

jameshughes89:
### What
Rework the introduction to the topic to include the problem solving strategy from the boolean topic. 

### Why
It feels like it belongs better under this topic, so this move (and rework) was done. Also, the compounding interest example isn't really a great example of building bigger functions since it legit can be one line long.

### DIFF
--- a/site/topics/car-rental/car-rental.rst
+++ b/site/topics/car-rental/car-rental.rst
@@ -14,10 +14,36 @@ What have we seen so far?
 * Logic
 * If/Else
 
+* Each of the individual topics may feel simple on their own
+* The difficult part tends to be when putting these topics together to solve complex problem
+
+
+Writing Bigger Programs
+=======================
+
+* There is no single correct way to write programs, but there are some strategies
+* For now I recomment a bottom up, incremental approach
+
+    # Start with an empty function and have it return some arbitrary constant value (e.g., ``0``)
+    # Run the function and verify that it does what you expect
+    # Add one or two lines of code
+    # Run the function and verify that it does what you expect
+    # Repeat
+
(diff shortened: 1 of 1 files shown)


site/topics/car-rental/car-rental.rst
```diff
@@ -14,11 +14,37 @@ What have we seen so far?
 * Logic
 * If/Else
 
+* Each of the individual topics may feel simple on their own
+* The difficult part tends to be when putting these topics together to solve complex problem
```
twentylemon, line 18: problem -> problems

site/topics/car-rental/car-rental.rst
```diff
@@ -14,11 +14,37 @@ What have we seen so far?
...
+* For now I recomment a bottom up, incremental approach
+
+    # Start with an empty function and have it return some arbitrary constant value (e.g., ``0``)
+    # Run the function and verify that it does what you expect
+    # Add one or two lines of code
+    # Run the function and verify that it does what you expect
+    # Repeat
+
+* This incremental strategy is great because a lot of the problem solving you will be doing will be incremental
+* ADdditionally, it helps you make sure everything is working along the way
```
twentylemon, line 34: ADdditionally -> Additionally 

Some kind of grammar checker would be nice to have as part of the build.

twentylemon (approved):

site/topics/car-rental/car-rental.rst
jameshughes89: I think maybe in the last pycharm update some plugin stopped working because I swear it was flagging these before.

PR #166, opened 2022-04-25
💣:memo: Remove Site From gitignore

jameshughes89:
### What
Remove `/site` from gitignore. 

### Why
So I can add stuff.

### Additional Notes
I was wondering why every new file I made was ignored.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -129,9 +129,6 @@ venv.bak/
 # Rope project settings
 .ropeproject
 
-# mkdocs documentation
-/site
-
 # mypy
 .mypy_cache/
 .dmypy.json


twentylemon:
`site` is the build output. It's meant to be a temporary store. Put files elsewhere and copy them into `site` as part of the build.

jameshughes89:
I believe, as of now, `build/` is the build output and `site/` is where all the `.rst` files are.

twentylemon:
It's actually [/out](https://github.com/jameshughes89/cs101/blob/main/setup.py#L19). Though the [deploy](https://github.com/jameshughes89/cs101/blob/main/.github/workflows/sphinx-build.yml#L60-L78) action treats it as `/site`. That probably needs to be fixed.

twentylemon:
Well. Maybe it's ok. I guess it'll just merge the two directories. Could lead to some confusion if stuff gets clobbered. but whatev.

twentylemon (approved):

jameshughes89:
Good catch on the `site` in the deploy. Options
1. Rename `site/` with all the rst to something else
2. Change deploy action to call it `out`

Number 2 is probably easier.

PR #168, opened 2022-04-25
:robot: Deploy action site -> builtsite

jameshughes89:
### Related Issues or PRs
As discussed in #166 

### What
Rename `site` to `builtsite`.

### Why
Eliminate potential confusion. 

### Testing
No, but also kinda' yes


twentylemon:
`out` seems more fitting, since it is the same `out` as the sphinx build uses. No need to introduce weird synonyms.

twentylemon (approved):

PR #170, opened 2022-04-26
🐛:lipstick: Correct Spelling

jameshughes89:
### What
Correct spelling throughout the topics so far. 

### Where
Up to what was already updated in the content. 

### How
Got the spellchecker working again. Turns out I had PyCharm in "Reader Mode" which I guess means it will not worry about spelling. 🤷


twentylemon (approved):

jameshughes89:
There are some github actions for spellcheck, but I also found [pyspelling](https://facelessuser.github.io/pyspelling/) that may be easier to work with.

PR #175, opened 2022-04-26
🎓 If/Else Topic Rework Intro

jameshughes89:
### What
Rework the intro to the if/else topic. 

### Testing
Yes.


site/topics/if-else/if-else.rst
```diff
@@ -1,45 +1,89 @@
...
 
-* We're pretty good at telling python *exactly what to do* now.
-* But there is no room for choice. It's just: *do these steps in this order*
-* That doesn't give us much flexibility (or computational power)
+    Using only what we have learned so far, write a function ``smush(a_number)`` that checks if the number is
+    positive or negative. If ``a_number`` is positive, the function will return half the value of ``a_number``. If
+    the number is not positive, simply return ``a_number``.
 
 
+Conditional Statements
```
twentylemon, line 12: statement -> expression. A statement is an action, like `if`/`else`, which may consist of several expressions. An expression, on the other hand, is something that evaluates to a value. A conditional expression is an expression which evaluates to a boolean value.

site/topics/if-else/if-else.rst
```diff
@@ -1,45 +1,89 @@
...
+        otherwise, return the value of a_number.
+
+        :rtype: float
+        :param a_number: Some arbitrary number.
+        :return: Half of a_number when it is positive, a_number when not positive.
+        """
+        return_value = a_number
+        if a_number > 0:
+            return_value = return_value / 2
+        return return_value
```
twentylemon, line 45: Stylistic choice, you should _always_ have an `else` block.

twentylemon (approved):

site/topics/if-else/if-else.rst
```diff
@@ -1,45 +1,89 @@
-**************
-Logic, if/else
-**************
+*******
+If/Else
+*******
 
-Conditionals
```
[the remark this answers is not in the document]
jameshughes89: Good catch my dude.

site/topics/if-else/if-else.rst
jameshughes89: The only reason I didn't here was because I'm hitting `else` in the next next section.

site/topics/if-else/if-else.rst
jameshughes89: I believe _technically_ all expressions are statements. Nevertheless, your point will be incorporated as, at the very least, is more correct.

site/topics/if-else/if-else.rst
twentylemon: Fair enough. I don't think `else` would add much weight here but it will definitely add at least a little weight (imo, it makes the code simpler, but I also have non-zero experience).

jameshughes89: Perhaps the best way to hit this is to write `smush_version_2` that uses `else`.

PR #181, opened 2022-04-28
🎓 If Else Topic Introduce `else`

jameshughes89:
### What
Show that `else` is a thing. 

### Why
It's cool. 

### Testing
Yes.

### Additional Notes
The *note* originally only was going to include version 2, but I opted to include version 3 as a way to demonstrate how the same functionality can be achieved with different implementations. It is perhaps a little contrived, but usually at this stage of the course they don't really have an appreciation of how there are literally an infinite ways to do anything. I figure, showing a few alternatives (despite how small of an example this is) may be helpful.

### DIFF
--- a/site/topics/if-else/if-else.rst
+++ b/site/topics/if-else/if-else.rst
@@ -193,9 +193,9 @@ Alternative Execution
             else:
                 return a_number
 
-    Another possibility is ``smush_version_3``. You will notice how similar it is to version 2, but here we simply
-    include the ``return``\s in the ``if`` and ``else`` blocks. Again, this is not *more correct* and it is only shown
-    here to demonstrate how the same functionality can be implemented differently. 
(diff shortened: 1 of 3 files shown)


site/topics/if-else/if-else.rst
```diff
@@ -101,44 +101,101 @@ Compound conditions
...
+            otherwise, return the value of a_number.
+
+            :rtype: float
+            :param a_number: Some arbitrary number.
+            :return: Half of a_number when it is positive, a_number when not positive.
+            """
+            if a_number > 0:
+                return a_number / 2
+            else:
+                return a_number
```
twentylemon, line 194: version 4; `return a_number / 2 if a_number > 0 else a_number`

twentylemon (approved):

site/topics/if-else/if-else.rst
jameshughes89: ew

PR #183, opened 2022-04-28
🎓 If Else Topic Nested Conditions

jameshughes89:
### What
Show example of nesting conditions. Show how it can also be done with `and`. 

Remove old stuff not getting focused on. 

### Additional Notes
Floating point precision came up in the stuff that was cut, so it will need to be hit somewhere.


site/topics/if-else/if-else.rst
```diff
@@ -173,114 +173,47 @@ Chains of alternatives
...
-* Again, no limit to how deep you nest... but mind the readability of your code!
-    * Actually, technically, Python has a limit of 20 nested statements, but other languages have much deeper limits
-    * In a more mathematical/theoreticaltical sense, there is no limit 
-
-* Do not go to anyone with this::
-
-	if a > 0:
-	   if b > 0:
-	      if c > 0:
-	         if d > 0:
```
twentylemon, line 207: Good riddance. Holy junk.

twentylemon (approved):

PR #184, opened 2022-04-28
💄 Change `\r\n` -> `\n`

jameshughes89:
### What
Make all EOLs `\n`

### Why
Better this way. Also git prefers it. 

### Where
Everywhere.

### Additional Notes
I left .gitignore alone because that one weird thing.


jdelamer:
It comes from Windows.
That's the only OS using \r\n for EOL.

twentylemon:
IIRC you can have git do this automatically if you have an appropriate `.gitattributes` in the repository root.

twentylemon (approved):

PR #197, opened 2022-05-02
🎓 Car Topic `num_kms_above_average`

jameshughes89:
### What
Write the section for `num_kms_above_average`

### Additional Notes
Ignore the name of the branch. It's lying.

### DIFF
--- a/site/topics/car-rental/car-rental.rst
+++ b/site/topics/car-rental/car-rental.rst
@@ -202,33 +202,54 @@ Average Kilometers Per Day
     assert 0.5 == average_kms_per_day(3, 1.5)
 
 
+Kilometers Above Allowable Average
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-Ok. Now for something harder... Number of kms over the daily average allowance. What do we know?
-    * Function to calculate the daily average 
+* Now for something a little harder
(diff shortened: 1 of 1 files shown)


twentylemon (approved):
Righto. I said this in satisfactory too -- consider starting at the end. In this case, try reversing the section order. Instead of laying foundational code like `total_kms` and this section, start with the next section -- tackle the problem immediately. Convert the problem statement to (broken) code, then implement the new functions you want to use.

In real world problems, you can't see the base components you want to use in order to solve the problem -- you can't just skip to the end of a design. Design is the process of breaking something apart so you can compose the pieces together, not finding ways to compose already exsting pieces.

jameshughes89:
> Righto. I said this in satisfactory too -- consider starting at the end. In this case, try reversing the section order. Instead of laying foundational code like `total_kms` and this section, start with the next section -- tackle the problem immediately. Convert the problem statement to (broken) code, then implement the new functions you want to use.
> 
> In real world problems, you can't see the base components you want to use in order to solve the problem -- you can't just skip to the end of a design. Design is the process of breaking something apart so you can compose the pieces together, not finding ways to compose already exsting pieces.

Pog.

I think it'll be easier to finish this and the next section and then do the reorder. With all the code there, I think the massaging will be simpler.

PR #198, opened 2022-05-03
🎓 Car Rental Topic Reorder and Finish

jameshughes89:
### Related Issues or PRs
Based on the suggestion from @twentylemon in #197 

### What
Reorder the way the ideas are presented such that we start at the _end_, work our way down, solve sub problems, and then work our way back to the end. 

Also finished off the topic. 

### Additional Notes
1. @jdelamer, you will need to update the colab link. 
2. I opted not to bother with the `assert`s for the input part and just did prints. There is nothing stopping the use of `assert` for the types.

### DIFF
--- a/site/topics/car-rental/car-rental.rst
+++ b/site/topics/car-rental/car-rental.rst
@@ -88,7 +88,11 @@ An Incremental Solution
 
 * Half of the description is IO
 * We know how to do this, so we will start here
+* The other half of the description is the calculation
 
+**Step 3**
+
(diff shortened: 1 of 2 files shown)


jdelamer:
I updated the colab file. So no need to update the link itself.

jameshughes89:
Removed the input stuff as it broke tests. I'm not too worried about removing it as the functions are the important bits and the input is presented in the topic anyway.

twentylemon (approved):
porg

PR #204, opened 2022-05-03
🎓 String Topic Mutability

jameshughes89:
### What
Say that strings are immutable 

### Why
Because they are

### DIFF
--- a/site/topics/strings/strings.rst
+++ b/site/topics/strings/strings.rst
@@ -88,23 +88,33 @@ More loops
 * But there is nothing wrong with the ``while``! There are just a lot of ways to solve a problem.
 * You can actually prove that there are a (countably) infinite number of programs to do any given task.
     * Some are just more efficient, and easier to read, than others.
-   
+
+
 Mutability
 ==========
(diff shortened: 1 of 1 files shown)


twentylemon (approved):

site/topics/strings/strings.rst
```diff
@@ -88,23 +88,33 @@ More loops
...
-        File "<stdin>", line 1, in <module>
-    TypeError: 'str' object does not support item assignment
-* Variables of some types (like ``int``) are *mutable*... that is: they can be changed.
-* Based on the above... do you think *strings* are mutable?
-* You can't change a string. You have to make a *new one* based on the old one. 
-    >>> new_a = a[:7] + 'X' + a[8:]
-    >>> print(new_a)
-    My littXe string
-   
+
```
twentylemon, line 95: maybe section title _Immutability_ instead

jameshughes89: I thought of that the moment after I pushed and figured _meh_. Seeing another agree, I GUESS I will change.

PR #209, opened 2022-05-04
🎓 💣 Loops Topic Remove Newton's Method and Finish Topic

jameshughes89:
### What
- Remove the Newton's Method section. 
- Rework the algo section
- Add sum even function to src and tests

### Why
- It was confusing --- I was trying to clean it up, but it just got worse
- Cleanup 
- Cleanup 

### Testing
Yes

### Additional Notes
The Newton's Method part, although cool, ended up being an issue for as long as I can remember. Probably should have removed it years ago. They tend to not see the forest (loop) through the trees (scary math).


site/topics/loops/loops.rst
```diff
@@ -185,68 +185,32 @@ Tracing Through A Function By Hand
...
-* An algorithm is a description of a series of steps to solve a problem.
-* Algorithms can be presented in natural language, but are easier to turn into a program when presented in a formal language.
-* Finding an algorithm to solve most problems is *very hard*. You can make a career, get tenure, make millions of dollars in patent licensing, etc., "just" by developing algorithms.
-* As programmers though, we usually leverage existing algorithms and other things to make our lives easier. We often won't be starting from scratch (although, right now you are...)
-* The two most important concepts you will learn in this course (or really, what a computer scientist spends years learning) are:
-    * **ALGORITHM**
-    * **DATA STRUCTURE**
-* So we're half done! (Just kidding)
+    * You can make a career out of coming up with algorithms
+    * There are certain open problems that, if you solve, will literally get you a million dollars and plenty of fame
```
twentylemon, line 200: [Citation needed](https://en.wikipedia.org/wiki/Millennium_Prize_Problems). Not sure if it makes sense to link directly to P vs NP and the Reimann hypothesis (hell, people with degrees still don't appreciate the latter), but I don't think you can just leave this sentence hanging like that.

twentylemon (approved):
Probably best to remove Newton's method stuff, for sure. Kinda relies on at least some understanding of calculus, which is also first year stuff. But, students here may not be in any math at all. So yeet.

site/topics/loops/loops.rst
jameshughes89: Added

PR #213, opened 2022-05-04
:bomb: Remove Machine Learning and Data Visualization

jameshughes89:
### What
Remove Machine Learning & Data Visualization topics

### Why
They're cool, but:
- 161 is going to hit some core topics a little harder (objects, testing, lists)
- 161 _was_ trying to attract students from other disciplines 
- With CSCI 128 being cleaned up, the ML and viz topics are better suited for that course as that is more of a "CS for everyone" course.


twentylemon (approved):
Makes sense.

PR #214, opened 2022-05-04
🎓 Split FileIO and Exceptions topics

jameshughes89:
### What
Split the FileIO and exceptions topics. 

### Why
- Can hit FileIO a little more 
- Exceptions tended to take away and distract from the fileIO part of the topic

### Additional Notes
The exceptions topic will **not** get fuller --- it'll be kept simple and high high level.


site/topics/fileIO/fileIO.rst
```diff
@@ -0,0 +1,347 @@
+**********************
```
twentylemon, line 1: Is this file a copy of the original? Should I be reviewing more closely?

jameshughes89: Nah, just a split at this stage. Will be reworked more later.

twentylemon (approved):

PR #217, opened 2022-05-05
🎓 Lists Topic split into lists and references topics

jameshughes89:
### What
Separate the lists and references topics. References will follow the lists topic as they will be used to get the point across. 

### Why
Information overload in one topic

### Additional Notes
I am thinking I will probably take some stuff from the `more-lists` topic and move that into the first lists. May end up removing `more-lists` depending on how much moves.


site/topics/references/references.rst
```diff
@@ -0,0 +1,516 @@
+******************
```
twentylemon, line 1: I assume this is like #214 where this is copy/paste. I have not reviewed the content closely.

twentylemon (approved):

site/topics/references/references.rst
jameshughes89: Bingo. 

In future PRs where this happens, I will make it more explicit that it is NOT a content change.

PR #223, opened 2022-05-06
🎓 Strings Topic objects and methods

jameshughes89:
### What
Briefly discuss objects and methods

### Why
Just as a little teaser so when we discuss references and objects, it won't be ALL new. 

### Additional Notes
Was trying to hit the sweet spot between simplicity and precision. Let me know if it's too much one way or the other. Even when we hit objects later in the course, things will need to be high-level with a number of asterisks.


site/topics/strings/strings.rst
```diff
@@ -403,99 +403,91 @@ Objects
...
+    * ``upper(a_string)``
 
-    * Not necessarily the best way, or a way you like
 
-4. There's also a good bookkeeping argument too
+* In reality, there is nothing stopping us from writing the function ``upper(a_string)``
+* But with the string, the sequence of characters that make up the string are stored in the object
 
-    * Put all the string methods with the strings
+    * The method ``a_string.upper()`` is modifying the data within the object
```
twentylemon, line 477: *Can* modify. Strings are immutable.

site/topics/strings/strings.rst
```diff
@@ -403,99 +403,91 @@ Objects
...
+    * The functionality that modifies the data in that object will belong to that object
```
twentylemon, line 478: Fun fact, this is actually a bad thing. 

I'll just leave that there.

twentylemon (approved):
Level seems fine to me for a brief introduction. Using methods seems like fair as an intro to objects.

site/topics/strings/strings.rst
jameshughes89: When I wrote that, I thought to myself "here's hoping no one says 'wait, you said they can't change'"

site/topics/strings/strings.rst
jameshughes89: I feel ya --- this is hitting the whole statefull thing. 

Any suggestions on a better presentation?

jameshughes89: I suppose it could just be _about_ the data in the object.

twentylemon: You need to teach objects unfortunately, that's just the state of programming languages. My completely unbacked point is asserting objects are a bad way to encapsulate data and functions -- they are best left to data and functions respectively.

PR #236, opened 2022-05-10
🎓 Lists Topic loops on lists

jameshughes89:
### What
Looping on lists

### How
Mostly examples

### Additional Notes
1. Not sure if this is the best place to introduce `range`
2. May want to move the subsequent sections to the topic *after* references
    * This topic is already getting long
    * It makes sense to talk about aliases after references
3. Need to go and check for indentation consistency throughout the lecture content since colab likes 2 spaces. This is a setting on colab that can be changed, but I only did it now and then realized the issue.

### DIFF
--- a/site/topics/lists/lists.rst
+++ b/site/topics/lists/lists.rst
@@ -115,28 +115,33 @@ Mutability
 Lists & Loops
 =============
 
-* ``for`` loops can be used to execute a block of code for every element in a list::
+* Similar to how ``for`` loops made it easy to iterate over each character in a string
+* ``for`` loops can be used the same way on lists to iterate over each element in the list
 
-    for element in some_list:
-        do_something(element)
+    * ``for`` each *thing* in a collection of *things*
 
-* Just like the loop we did with Strings last class!
-* This is incredibly useful. In fact, you've already seen it in assignment 1. Let's try it::
+.. code-block:: python
+    :linenos:
 
-    def like_food(food_list):
-        for food in food_list:
-            if food not in ['McDonalds','Burger King']:
-                print('I like ' + food)
-            else:
-                print('I dont like ' + food + ' so much.')
+    collection_of_things = ["Hello", 10, True, 100.001]
 
-* And now we'll run our function:
+    for thing in collection_of_things:
+      print(thing)
 
-    >>> like_food(['curry','sushi','McDonalds','bison'])
-    I like curry
-    I like sushi
-    I dont like McDonalds so much.
-    I like bison
+    # Results in
+    #   Hello
+    #   10
+    #   True
+    #   100.001
+
+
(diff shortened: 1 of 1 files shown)


site/topics/lists/lists.rst
```diff
@@ -112,13 +112,173 @@ Mutability
...
+* Using the ``range`` function would help in this situation as we can use it to loop over each index in ``haystack``
+* The only catch is that we need to actually index ``haystack``
+
+.. code-block:: python
+    :linenos:
+
+    def index_of(needle, haystack):
+        haystack_length = len(haystack)
+        for i in range(haystack_length):
+            if haystack[i] == needle:
```
twentylemon, line 229: [enumerate](https://docs.python.org/3/library/functions.html#enumerate) is specifically for this case.

site/topics/lists/lists.rst
```diff
@@ -112,13 +112,173 @@ Mutability
...
+    #   3 d
+    #   4 e
+
+
+* Here there is no need to actually index the list since the ``element`` variable already has the value ``another_list[i]``
+
+.. note::
+
+    Some people love using ``enumerate``, others feel it is unnecessary since they do not see indexing the list as an
+    imposition. Neither is more correct than the other --- use what you are comfortable with.
```
twentylemon, line 274: Well. The argument is intent. The goal is to loop through the list, adding a layer of indirection through `range` is needlessly complicating. _Simplicity_ is the argument -- I'd argue that is indeed _more correct_.

jameshughes89: perhaps _neither is wrong_.

jameshughes89: The argument _against_ using it here, in intro CS, is avoiding another keyword that is particularly python-ie.

twentylemon: It's a function, not a keyword. But maybe fair enough. Probably too early to introduce tuples and multiple returns.

I'd still maybe change the wording here. `enumerate` is more appropriate, but delves into some python features you have not (and will not?) touched on in the course.

jameshughes89: Oh, how about, yeet the "note" in question?

twentylemon: Eh. Also fine.

site/topics/lists/lists.rst
```diff
@@ -268,11 +268,6 @@ Looping Over Indices and Enumerate
 
 * Here there is no need to actually index the list since the ``element`` variable already has the value ``another_list[i]``
 
-.. note::
-
-    Some people love using ``enumerate``, others feel it is unnecessary since they do not see indexing the list as an
-    imposition. Neither is more correct than the other --- use what you are comfortable with.
-
```
jameshughes89: Note cut

PR #241, opened 2022-05-12
🎓 References Topic references

jameshughes89:
### What
Explain references


site/topics/references/references.rst
```diff
@@ -96,63 +96,62 @@ Lists in Memory
...
 
->>> w[4] = P
+.. note::
 
-.. image:: array_pointer_copy_change.png
+    As mentioned earlier, in reality Python would not actually store the contents of the list within the block of memory
+    allocated for the list. Instead, Python actually stores the contents of the list in their own memory locations and
+    the memory location for the list stores references to the location of the respective contents. This detail is
+    important, but fortunately we could ignore this detail here for the purpose of learning about references. Further,
+    Python takes care of these details behind the scenes so we do not need to think about this too much.
```
twentylemon, line 150: This is a bit wordy. Saying more simply that a list is itself also a reference to that array (among other details, like the length) would be sufficient here. References were just explained, I think it's fair enough to fall back on that, especially for something that is more of an in-passing note.

twentylemon (approved):

PR #245, opened 2022-05-16
🎓 Tuples, Dictionaries, and Sets Topic discuss tuples 

jameshughes89:
### What
Show and motivate tuples

### Additional Notes
I was considering including examples of:
- `enumerate` returning a tuple and how unpacking works
- multiple values getting returned form a function
- `zip` to show packing

but I opted to not include this since:
- I do not want to distract from the main core ideas of the course
- It may be too much info they worry about learning when they probably don't use them again during the semester
- If/when the time comes that they discover multiple returns, etc., being aware of tuples will probably be enough to connect the dots. 

All that said, I am not overly committed one way or the other so if there is a good argument to include these things, let me know.


site/topics/dictionaries-sets/dictionaries-sets.rst
```diff
@@ -2,61 +2,56 @@
...
+        print(some_tuple[0])   # Results in 10
 
-    >>> coord = (1, 2)
 
-* So above we're storing it as a tuple. Nothing special so far here either.
+* But unlike lists, tuples are immutable
+
+    * Once they are created, we cannot change them
+
+* When compared to lists, tuples are quite *light weight*
```
twentylemon, line 37: I don't really see performance being a great argument for tuples in the context of this course. I'd stick to "these exist and can be useful, you may come across them in python."

jameshughes89:
Also renamed directory & file as tuples were not cut as originally thought

site/topics/tuples-dictionaries-sets/tuples-dictionaries-sets.rst
```diff
@@ -2,61 +2,50 @@
...
+
+    * For example, a tuple would be great for storing cartesian coordinate ``(x, y)``
+    * Tuples were also used in the Starbucks assignment to store the latitude and longitude pairs
+
+    .. code-block:: python
+        :linenos:
+
+        for row in starbucks_file_reader:
+            location_tuple = (float(row[0]), float(row[1]))
+            starbucks_locations.append(location_tuple)
```
twentylemon, line 47: > * `enumerate` returning a tuple and how unpacking works

`enumerate` might be a better motivating example. Though I'm not a fan of using tuples in interfaces.

twentylemon (approved):

PR #254, opened 2022-05-20
🎓 Dictionaries 

jameshughes89:
### What
Explain dictionaries, but at a pretty high level. 

### Why
They're awesome, but the nitty gritty details are outside the scope of the course. 

Further, it is expected that, although they will [likely] use dictionaries in A4, they will not really remember them or make much use of them. The main purpose of showing them dictionaries is because they will start making more use of them in the following course.

### DIFF
--- a/site/topics/tuples-dictionaries-sets/tuples-dictionaries-sets.rst
+++ b/site/topics/tuples-dictionaries-sets/tuples-dictionaries-sets.rst
@@ -11,7 +11,7 @@ Tuples
 
 * A tuple looks and behaves similar to a list, but has slightly different syntax
 
-    * Tuples use parentheses ``(``\, ``)``)
+    * Tuples use parentheses ``(``\, ``)``
 
     .. code-block:: python
         :linenos:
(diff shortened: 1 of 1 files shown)


site/topics/tuples-dictionaries-sets/tuples-dictionaries-sets.rst
```diff
@@ -49,122 +49,112 @@ Tuples
...
+    :linenos:
+
+    some_dictionary["Sally"] = 90
+    print(some_dictionary["Sally"])         # Results in 90
+
 
-    # Creates some lookup table list thing
-    grades = [['James', 98], ['Bob', 86], ['Janice', 86], ['Greg', 59]]
+Why They Are Great
+------------------
```
twentylemon, line 104: This hardly does it justice, but I guess it'll suffice. I'm not really sure how to say "maps are _the_ primary data structure" and have that carry any meaning in this context.

twentylemon (approved):

PR #258, opened 2022-05-24
🎓 Debugging syntax errors and type errors

jameshughes89:
### What
Show some of the errors they are likely to see error messages for. Only the syntax and type errors are discussed in their own sections, but time will be taken during lecture to hit each of those in the point form list. 

### Why
At this stage they have likely seen many of these errors, and may already be familiar with addressing them. Either way, it will help make sure people are on the same page-ish. 

### How
I do like my segue in Type Errors from error messages to no error message, and how the following section hits logic errors. 

### Additional Notes
This topic is going to be far far far from thorough. It's purpose is just to get them thinking about these things explicitly and hopefully they are equipped enough to learn more as they go.

### DIFF
--- a/site/topics/debug/debug.rst
+++ b/site/topics/debug/debug.rst
@@ -86,8 +86,8 @@ Type Errors
 
 * As we have seen, Python is pretty good about figuring out types
 
-.. admonition:: Activity
-    :class: activity
+.. code-block:: python
+    :linenos:
 
     a = 5           # It's an integer
     b = 5.5         # Float since there is a decimal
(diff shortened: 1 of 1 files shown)


site/topics/debug/debug.rst
```diff
@@ -28,60 +28,107 @@ The Importance of Debugging
...
-* As we've seen many times, Python is pretty good at transparently guessing how to change types when you ask it do something that involves multiple types::
-  
-    >>> 2.03 + 4
-    6.0299999999999994
-     
-* **But** sometimes you might ask the impossible:
 
-     >>> 2.03 + 'octopus'
-     TypeError: unsupported operand type(s) for +: 'float' and 'str'
+Type Errors
```
twentylemon, line 83: Think you're missing a new line after the code block. Syntax highlighting is outta whack.

twentylemon (approved):

PR #261, opened 2022-05-25
:mortar_board: Debugging delta debugging and rubber duck debugging

jameshughes89:
### What
Explain "delta debugging" (but not really) and rubber duck debugging

### Additional Notes
Next year the students will all be getting a StFX Branded Rubber Duck, so that's awesome.


twentylemon:
> Next year the students will all be getting a StFX Branded Rubber Duck, so that's awesome.

I want.

PR #263, opened 2022-05-26
🎓 File io CSV files

jameshughes89:
### What
Reading and writing CSV files

### Why
They're greeeeat

### DIFF
--- a/site/topics/fileIO/fileIO.rst
+++ b/site/topics/fileIO/fileIO.rst
@@ -188,10 +188,9 @@ Writing to a CSV File
 * In the above example, notice that all the data for the row is contained within a list
 
 
-
-For next class
+For Next Class
 ==============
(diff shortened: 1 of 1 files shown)


twentylemon (approved):
The activities in the IO parts so far have seemed like larger conceptual jumps compared to previous ones. The code examples are nice, but explanations would be beneficial. Lecture accompaniment is sufficient, I think, but including it in the notes serves at least as a reminder to yourself, which is always nice.

jameshughes89:
> The activities in the IO parts so far have seemed like larger conceptual jumps compared to previous ones. The code examples are nice, but explanations would be beneficial. Lecture accompaniment is sufficient, I think, but including it in the notes serves at least as a reminder to yourself, which is always nice.

:+1:

PR #275, opened 2022-06-01
🎓 Exceptions example exit()

jameshughes89:
### What
Example of the program halting once the exception is seen

### DIFF
--- a/site/topics/exceptions/exceptions.rst
+++ b/site/topics/exceptions/exceptions.rst
@@ -149,18 +149,21 @@ Bob example
 	  
 	  
 	  
-Tim example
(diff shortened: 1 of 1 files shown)


twentylemon (commented):
I don't really see this as an effective way of handling the exception. Letting it bubble up is honestly preferable to just exiting.

jameshughes89:
> I don't really see this as an effective way of handling the exception. Letting it bubble up is honestly preferable to just exiting.

I ended up adding an "alternative" (what you suggested) as a way to also bring up the fact that they propagate up.

twentylemon (approved):

PR #278, opened 2022-06-02
🎓 Circle class

jameshughes89:
### What
Create a `Circle` class for introducing objects. 

### Why
- It's simple in that it has one attribute.
- The methods are simple and need gr8 math
- The methods are more data-_ie_

### Testing
:+1:

But also, around here may be a good point to start to introduce `unittest` since we're hitting (a) classes and (b) tests are becoming more than just an `assert` in that they may need some setup. 

### Additional Notes
I figure
1. Circle in OBJECTS topic 1 to introduce the idea. Also hit attributes and simple methods. 
2. Next topic will be `Point`, but will have a chance to make methods more complex
3. Next next topic will be a collection, so like a `FriendList` or something, idk. Specifics are TBD at this point. 
4. We all know how _someone_ feels about objects 😉.

### DIFF
--- a/src/circle.py
+++ b/src/circle.py
@@ -15,3 +15,16 @@ class Circle:
     def circumference(self) -> float:
         return 2 * math.pi * self.radius
 
+
+circle_zero = Circle(0)
+assert 0 == circle_zero.radius
+assert 0 == circle_zero.diameter()
+assert 0 == circle_zero.area()
+assert 0 == circle_zero.circumference()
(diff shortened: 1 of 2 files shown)


twentylemon:
> 4\. We all know how _someone_ feels about objects wink.

It's not a feeling so much as an innate truth.

twentylemon (approved):
fine sure whatever

PR #280, opened 2022-06-03
🎓 Point class

jameshughes89:
### What
A Point class for a topic covering more on objects. 

### Why
The 2nd objects topic will focus more on methods and the point class has more methods that seem a little "fancier" than the circle (parameters, returning something of the same type).

### Testing
:+1:, though, I am longing for parameterized tests. 

### Additional Notes
I'm really liking the idea of introducing `unittest` after this topic such that the topics go:
- Objects 1 --- constructor and attributes and methods
- Objects 2 --- More on methods
- Unit testing with `unittest` 
- Objects 3 --- Collection (list of some objects) 

Motivation for the `unittest` topic is:
1. Asserts are starting to be limiting as the tests are requiring a little more setup
2. Asserts get jumbled together and not trivial to understand what the test is just by looking at it 
3. If the first assert fails, we stop the whole thing, which is sad

The only thing I don't love is the "magic" code --- "TestCircle(*unittest.TestCase*):", but that shouldn't be too hard to address.

### DIFF
--- a/src/point.py
+++ b/src/point.py
@@ -2,7 +2,7 @@ import math
 
 class Point:
 
-    def __init__(self, x=0, y=0):
+    def __init__(self, x, y):
         self.x = x
         self.y = y
 
--- a/src/point.py
+++ b/src/point.py
@@ -28,7 +28,7 @@ class Point:
         :return: The Euclidean distance between the self Point and the parameter Point
         :rtype: float
         """
-        return math.sqrt((self.x - point.x)**2 + (self.y - point.y)**2)
+        return math.sqrt((self.x - point.x) ** 2 + (self.y - point.y) ** 2)
 
     def find_midpoint(self, point):
         """
@@ -39,8 +39,8 @@ class Point:
         :return: A Point at the midpoint between the self Point and the parameter Point
         :rtype: Point
         """
-        midpoint_x = (self.x - point.x)/2 + point.x
-        midpoint_y = (self.y - point.y)/2 + point.y
+        midpoint_x = (self.x - point.x) / 2 + point.x
+        midpoint_y = (self.y - point.y) / 2 + point.y
         return Point(midpoint_x, midpoint_y)
 
     def __eq__(self, other) -> bool:
--- a/src/point.py
+++ b/src/point.py
@@ -1,9 +1,23 @@
 import math
 
+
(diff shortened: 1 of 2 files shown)


test/test_point.py
```diff
@@ -0,0 +1,91 @@
+import unittest
+
+from src.point import Point
+
+
+class TestCircle(unittest.TestCase):
```
twentylemon, line 6: `Point`, not circle.

Kinda aside, it is more typical that test class names end with `Test` (ie, testing `ClassName` in `ClassNameTest`), rather than start with them. `unittest` doesn't actually care, you're free to define your own style. I've also seen the prefix style used (usually in older projects), but my preference is the suffix, since it's more like how you'd just speak about it. X tests are in the XTest class, rather than TestX class.

src/point.py
```diff
@@ -0,0 +1,94 @@
...
+        self.y = y
+
+    def distance_from_origin(self) -> float:
+        """
+        Calculate the Euclidean distance from this Point (self) and the origin (0, 0).
+
+        :return: The Euclidean distance from the self Point and the origin
+        :rtype: float
+        """
+        return math.sqrt(self.x**2 + self.y**2)
```
twentylemon, line 20: Might be more instructive to `return self.distance_from_point(Point(0,0))`

twentylemon (approved):

src/point.py
jameshughes89: I went back and forth on that one, but seeing a 2nd-er for the idea, I'm game.

test/test_point.py
jameshughes89: [I went with what `unittest` showed](https://docs.python.org/3/library/unittest.html#basic-example), but I am not particularly committed to either way.

jameshughes89: (╯°□°)╯︵ ┻━┻

jameshughes89: I suppose an argument for keeping it as `TestX` is the naming convention of the file names themselves, as they have `test_` at the beginning of file names.

twentylemon: I'd say that's an argument for `XTest`. Functions are actions -- `test_this_thing`. Namespaces are nouns -- `ThingTest`.

PR #286, opened 2022-06-06
💣 Lab debugging remove lab ✂️ 

jameshughes89:
### What
Remove the debugging lab

### Why
1. Kinda' silly
2.  Added another lab on lists, so now there are too many
3. Feels like they probably wont need me to give them a lab to see broken code 

**Content is kept in repo in case it is still needed, but the link to the lab was cut**


twentylemon (approved):
You _could_ opt to leave the lab file in the repo. If you ever resurrected it, you wouldn't have to dig through source control or remember that you had one.

jameshughes89:
> You _could_ opt to leave the lab file in the repo. If you ever resurrected it, you wouldn't have to dig through source control or remember that you had one.

The reason I will do this and keep the rst file is because, as I started working on the other labs, I was a little worried if the pacing of the course content would be off and I would need another lab before the Objects.

PR #290, opened 2022-06-06
🎓 Course class

jameshughes89:
### What
A `Course` class for keeping track of students. 

### Why
They get to make a collection. 

### Testing
:+1:, and yeah --- will defo have to have covered `unittest` before this one.

### DIFF
--- a/src/Course.py
+++ b/src/Course.py
@@ -1,0 +1,11 @@
+class Course:
+    """
+    A collection of students enrolled in a course. This class manages the individual Students and provides simple
+    enrollment details.
+    """
+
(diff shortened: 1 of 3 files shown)


jameshughes89:
Also, I am deliberately not including any more of the magic methods because:
1. I'm not quite sure why, but the students are thrown off by them a little. Maybe it's just too much all at once?
2. If/when they figure out the magic methods, it will give them another potential thing to complain about in Java "PyThOn Is BeTtEr ThAn JaVa BeCaUsE ...."

twentylemon (approved):
My tongue hurts.

PR #293, opened 2022-06-07
🎓 Change Point class to Point3D class

jameshughes89:
### What
Make the `Point` class a `Point3D` class. This just adds a dimension. 

### Why
I want to take the idea that the book has on the latter class topic where the shape has a point defining it's location in space. The book example used a Rectangle, but I will do a Sphere. 

The Sphere class will be:
- attributes (centre, radius)
- vol, surface area
- method to check if the spheres overlap 
- make a new sphere between the 2 with a specified radius (this is easier than doing a rectangle)
- methods to check distance between centre points and another for distances between edges (probably left as exercises for the students though)

The motivation of this is to have a class containing another class and show how we can use the `Point3D` level of abstraction to help with making the sphere class. 

### Testing
:+1: Changed them over to `Point3D` and re-ran


src/point3d.py
```diff
@@ -0,0 +1,96 @@
...
+    def __repr__(self) -> str:
+        """
+        Generate and return a string representation of the Point3D object.
+
+        This os a "magic method" that can be used with `str(some_point3d)` or for printing.
+
+        :return: A string representation of the Point3D
+        :rtype: string
+        """
+        return f"Point({self.x}, {self.y}, {self.z})"
```
twentylemon, line 73: `Point3D`

test/test_point.py
```diff
@@ -1,72 +1,81 @@
...
+        point = Point3D(0, 0, 0)
+        self.assertEqual(0, point.z)
+
     def test_update_point_x_has_updated_x(self):
-        point = Point(0, 0)
+        point = Point3D(0, 0, 0)
         point.x = -1
         self.assertEqual(-1, point.x)
 
     def test_update_point_y_has_updated_y(self):
```
twentylemon, line 24: What are you testing with these? This is basically a language feature.

test/test_point.py
```diff
@@ -1,72 +1,81 @@
 import unittest
```
twentylemon, line 1: File name needs updating

twentylemon (approved):

test/test_point.py
jameshughes89: Original thought --- _overkill will be good for having them think about testing thorough_ 
now --- _no, that's bad_

They were yeeted with the latter class tests, idk why I left them here 🤷

PR #294, opened 2022-06-07
📝 Add new lines back to template 🔬 

jameshughes89:
### Related Issues or PRs

Undoes #231 

### What
Add the new line back to the template files. 

### Why
It was getting annoying to discard the changes to the md files every time I ran format. mdformat insists on the new line, and I looked into adding an exception to that rule for these files, but whatever, it's easier to make it shutup this way. 

Also, if the good people of _mdformat_ believe the new line is important, I will yield to them.


twentylemon (approved):
Today is the beginning of the rest of your life. Formatters are good enough, there's better things to care about.

PR #295, opened 2022-06-07
🎓 Sphere class

jameshughes89:
### What

Sphere class that makes use of a `Point3D`

### Testing

:+1:
But you will notice a lack of `assert` tests. This has the benefit of
- This topic leads into the unittest topic
- I don't have to write them

### DIFF
--- a/src/sphere.py
+++ b/src/sphere.py
@@ -34,7 +34,7 @@ class Sphere:
         """
         return self.centre_point.distance_from_point(other.centre_point)
 
-    def distance_between_edges(self, other: Point3D) -> float:
+    def distance_between_edges(self, other: "Sphere") -> float:
         """
         Calculate and return the distance between the edges of two Spheres. If the value is negative, the two Spheres
         overlap.
@@ -46,7 +46,7 @@ class Sphere:
         """
         return self.distance_between_centres(other) - self.radius - other.radius
 
-    def overlaps(self, other: Point3D) -> bool:
+    def overlaps(self, other: "Sphere") -> bool:
         """
         Determine if two Sphere objects overlap within the 3D space. Two Spheres that are touching (distance of 0
         between edges) are considered overlapping.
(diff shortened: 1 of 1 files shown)


src/sphere.py
```diff
@@ -0,0 +1,63 @@
+import math
+
+
+class Sphere:
+    """
+    Class for managing Spheres within a 3D space.
+    """
+
+    def __init__(self, centre_point, radius):
```
twentylemon, line 9: Why no type hints?

twentylemon (approved):

src/sphere.py
jameshughes89: I added the ones I could, buy adding more requires `typing`. That said, if I am introducing `unittest`, I could do `typing` with it like I did in the original testing and type hints topic early in the course.

jameshughes89: `float` was added, but to use `Point3D`, I need typing

```python
from typing import Type
...
    def __init__(self, centre_point: Type[Point3D], radius: float):
```

twentylemon: [nope](https://github.com/duck-dynasty/duckbot/blob/main/duckbot/slash/slash_command_handler.py#L17)

jameshughes89: my life has been a lie

twentylemon: [Type[T]](https://docs.python.org/3/library/typing.html#typing.Type) is for generic types. Classes, you can just use use the class name.

jameshughes89: Gonna' go back and add type hints to previous classes now.

src/sphere.py
```diff
@@ -0,0 +1,67 @@
...
+    def diameter(self) -> float:
+        return 2 * self.radius
+
+    def surface_area(self) -> float:
+        return 4 * math.pi * self.radius**2
+
+    def volume(self) -> float:
+        return (4 / 3) * math.pi * self.radius**3
+
+    def distance_between_centres(self, other: Point3D) -> float:
```
twentylemon, line 26: `other` is presumably a sphere, not a point

PR #296, opened 2022-06-07
:new: Add type hints to existing classes

jameshughes89:
### Related Issues or PRs
As mentioned in #295 by @twentylemon, turns out custom classes do not require `typing` 

### What

Add type hints to the classes made. 

### Where

All the classes made so far 

### How

In some cases the type hints needed to be added as string literals since they were of the class that was being defined, otherwise they were just added. I believe there is 

### Testing

:+1: for good measure.

### DIFF
--- a/src/circle.py
+++ b/src/circle.py
@@ -42,7 +42,7 @@ class Circle:
         """
         return 2 * math.pi * self.radius
 
-    def __eq__(self, other) -> bool:
+    def __eq__(self, other: "Circle") -> bool:
         """
         Check if the Circle (self) is equal to the parameter Circle other. Circles are equal if they have the same
         radius.
--- a/src/point3d.py
+++ b/src/point3d.py
@@ -6,7 +6,7 @@ class Point3D:
     A class for representing points in a three dimensional (3D) space.
     """
 
-    def __init__(self, x, y, z):
+    def __init__(self, x: float, y: float, z: float):
         self.x = x
         self.y = y
         self.z = z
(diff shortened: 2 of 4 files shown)


src/circle.py
```diff
@@ -42,7 +42,7 @@ def circumference(self) -> float:
         """
         return 2 * math.pi * self.radius
 
-    def __eq__(self, other) -> bool:
+    def __eq__(self, other: "Circle") -> bool:
```
twentylemon, line 45: This is not actually true. `other` is `Any`

jameshughes89: Good catch. I'll have to remove the typehints in the equals as `Any` needs `typing`.

src/point3d.py
```diff
@@ -20,7 +20,7 @@ def distance_from_origin(self) -> float:
         """
         return self.distance_from_point(Point3D(0, 0, 0))
 
-    def distance_from_point(self, other) -> float:
+    def distance_from_point(self, other: "Point3D") -> float:
```
twentylemon, line 23: Self type hinting is usually done via an import.

```python
from __future__ import annotations

class A:
    def f(a: A):
        pass
```

I'm not sure if `"Point3D"` is the same, but I'd guess not.

jameshughes89: This was not done due to the import and overhead. It appears that a simple, and workable alternative is the string literal. 

Things fortunately seem smart enough to connect the dots. 
image

jameshughes89: [Also, reference to the PRP on it.](https://peps.python.org/pep-0484/#forward-references)

twentylemon (approved):

PR #300, opened 2022-06-07
🎓 Objects Topic 2 rename 🔬 

jameshughes89:
### What

Rename the 2nd Objects topic to "Interacting Objects" from "More Methods". 

### Why

1) Didn't like the name of it
2) Although the fancier methods will be part of the topic, the topic will have an emphasis on the objects interacting a little (point on point and points in circles, and circles with circles)


site/index.rst
```diff
@@ -135,7 +135,7 @@ Course Notes
     topics/fileIO/fileIO
     topics/exceptions/exceptions
     topics/objects/introduction
-    topics/objects/methods
+    topics/objects/interacting
```
twentylemon, line 138: I do kinda think methods fits better. /shrug

twentylemon (approved):

site/index.rst
jameshughes89: 🤔

jameshughes89: Given that "Methods" will probably be more helpful when doing a quick look at the table of contents, I agree --- Methods is perhaps better. Sections can be more precise about interacting objects.

PR #306, opened 2022-06-07
:new: Unittest Topic next class link to repo 🔬 

jameshughes89:
### What

Add a link to the GitHub repo so they can check out the unittests written for the course.

### DIFF
--- a/site/topics/testing/unittest.rst
+++ b/site/topics/testing/unittest.rst
@@ -45,4 +45,5 @@ Running Unit Tests
 For Next Class
 ==============
 
+* Check out the test folder in the `GitHub repo <https://github.com/jameshughes89/cs101>`_  to see the unit tests written for the course content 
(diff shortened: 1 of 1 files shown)


site/topics/testing/unittest.rst
```diff
@@ -45,4 +45,5 @@ Running Unit Tests
 For Next Class
 ==============
 
+* Check out the test folder in the `GitHub repo <https://github.com/jameshughes89/cs101>`_  to see the unit tests written for the course content
```
twentylemon, line 48: Maybe worth linking directly to the tests directory? Unless the intention is also to say "contribute, dammit."

twentylemon (approved):

site/topics/testing/unittest.rst
jameshughes89: It was originally a link right to the tests, but figured, _eh, maybe they'll poke around a little, which would be good_.

PR #312, opened 2022-06-08
💄 Change test class name order

### LINKED ISSUE #281: 💄  Rename test classes to be XTest instead of TestX
### Related Issues or PRs
As brought up in #280 by @twentylemon 

### What
Consider updating the unittest names to be XTest instead of the current TestX. Eg. `TestCircle` -> `CircleTest`.

### Why
Although it doesn't _realllllly_ matter, it does kinda flow more naturally in English. The argument _against_ this is that `test_` starts the test file names. 

### Where
All unittest test classes. 

### Additional Notes
Putting this in an issue so the idea does not get lost along the way.

jameshughes89:
### Related Issues or PRs
closes #281

### What

Change the test names to be XTest instead of TestX

### Why

Although there appears to be no "standard", @twentylemon had some method to the madness that made sense, so that's the idea. 

### Testing

:+1:

### DIFF
--- a/test/test_circle.py
+++ b/test/test_circle.py
@@ -3,7 +3,7 @@ import unittest
 from src.circle import Circle
 
 
-class TestCircle(unittest.TestCase):
+class CircleTest(unittest.TestCase):
     def setUp(self):
         self.circle_0 = Circle(0)
         self.circle_10 = Circle(10)
--- a/test/test_course.py
+++ b/test/test_course.py
@@ -4,7 +4,7 @@ from src.course import Course
 from src.student import Student
 
 
-class TestCourse(TestCase):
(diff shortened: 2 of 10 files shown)


twentylemon (approved):
I also prefer to name files `thingy_test.py`, but I don't remember if that's support by unittest. Might only be a pytest thing.

jameshughes89:
> I also prefer to name files `thingy_test.py` (to match the class name), but I don't remember if that's support by unittest. Might only be a pytest thing.

I believe `unittest` requires them to be `test_` (I had assumed it did not matter, but this was the source of _why tf are the tests not working?_ for a less than trivial amount of time)

PR #316, opened 2022-06-08
🎓 Objects Interacting Topic  `Point3D` class' methods 

jameshughes89:
### What

Discuss the methods within the `Point3D` class. 

### Additional Notes

Despite being within the "Methods" subsection, the "Magic Methods" subsubsection is skipped here as that would make this PR too big.

### DIFF
--- a/site/topics/objects/interacting-objects.rst
+++ b/site/topics/objects/interacting-objects.rst
@@ -30,6 +30,137 @@ Constructor and Attributes
 Methods
 -------
 
+* The methods we want are
+
+    * A way to measure the distance from the origin --- ``distance_from_origin``
+    * A way to measure the distance from another point  --- ``distance_from_point``
+    * A way to find the midpoint between two points --- ``find_midpoint``
+    * An ``__eq__`` method for equality
+    * A ``__repr__`` for generating a string representation of the object
+
+.. code-block:: python
+    :linenos:
+
+    class Point3D:
+
+        # init and/or other methods not shown for brevity
+
+        def distance_from_point(self, other: "Point3D") -> float:
+            """
+            Calculate the Euclidean distance from this Point3D (self) and the Point3D passed as a parameter.
+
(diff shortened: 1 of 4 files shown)


site/topics/objects/interacting-objects.rst
```diff
@@ -30,6 +30,67 @@ Constructor and Attributes
 Methods
 -------
 
+* The methods we want are
+
+    * A way to measure the distance from the origin --- ``distance_from_origin``
+    * A way to measure the distance from another point  --- ``distance_from_point``
+    * A way to find the midpoint between two points --- ``find_midpoint``
```
twentylemon, line 37: Does the sphere use this? Might as well not include it if it is not used.

twentylemon (approved):

site/topics/objects/interacting-objects.rst
jameshughes89: I'm conflicted on this one:
1. I agree --- it's not used, so one is left wondering _...k..._ 
2. I want more methods being written for the sake of examples
3. I like that the method shows the return of a new instance of the same class

Solutions:
(a) Remove method
(b) Include it and plug nose
(c) Include it somehow into the `Sphere` class --- maybe making a new sphere at the midpoint? Like fill the void between 2 spheres if they don't overlap, and raise an exception otherwise. 

Given points 1, 2, and 3, I am inclined to go with (c); but due to the secret point 4, which is _but I don't want to do more work here_, I am feeling _bleh_. 

Thoughts?

twentylemon: It's not used, don't include it. [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it). If the goal is to say "btw you can return instances of this class," you can throw that in a note, or jam it into an assignment. Finding the midpoint between two spheres is a kludge that has no meaningful purpose.

jameshughes89: image
Nice

PR #319, opened 2022-06-09
:new: Objects Topic Shuffle make new topic and move other topics around

jameshughes89:
### Related Issues or PRs

As discussed with @twentylemon  in #315 

### What

1. Make a new objects topic to come after the `Sphere` implementation
2. Introduce the topic
3. Move the Data structures topic back

### Why

1. The flow of the content will be improved

### Additional Notes

The existing PRs (#315, #316, #317, #318) will be updated by moving the content to this topic after this PR is merged.

### DIFF
--- a/site/topics/objects/interacting-objects.rst
+++ b/site/topics/objects/interacting-objects.rst
@@ -12,6 +12,8 @@ Objects III --- Interacting Objects
 
 * Although there is nothing wrong with the original ``Sphere`` implementation, let's consider that
 
+    * It will make our ``Sphere`` class simpler and easier to understand
+    * It allows us to write the methods at a higher-level of abstraction since can deal with ``Point3D`` objects instead of individual coordinates
     * If we wanted to make other three dimensional shapes, they could use something like a ``Point3D`` class to represent their position
     * It allows us to abstract and delegate some functionality to another class
     * It will help with testing since we can test the ``Point3D`` class' functionality and the ``Sphere``\s independently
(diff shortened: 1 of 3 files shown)


site/topics/objects/interacting-objects.rst
```diff
@@ -0,0 +1,63 @@
...
+
+    * Location in three dimensional space
+    * Distance between points
+    * Equality on points
+
+* Although there is nothing wrong with the original ``Sphere`` implementation, let's consider that
+
+    * If we wanted to make other three dimensional shapes, they could use something like a ``Point3D`` class to represent their position
+    * It allows us to abstract and delegate some functionality to another class
+    * It will help with testing since we can test the ``Point3D`` class' functionality and the ``Sphere``\s independently
```
twentylemon, line 17: It also makes `Sphere` smaller and easier to understand. You get to define methods based on  further abstractions, so there's less to think about. Checking if sphere's overlap, for example, can be checking if the distance between their centers is less than the sum of the radii -- ok, write that in code, no cruft about _how_ to calculate distance, the function you call worries about those details.

It can also allow for easier refactoring in the future. For example, you could change sphere to hypersphere by replacing `Point3D` with a generic n-dimensional `Point`. Then `Hypersphere` just works.

twentylemon (approved):

site/topics/objects/interacting-objects.rst
jameshughes89: > It can also allow for easier refactoring in the future. For example, you could change sphere to hypersphere by replacing `Point3D` with a generic n-dimensional `Point`. Then `Hypersphere` just works.

I love this point. I wont add it to the intro bit of the topic, but will hit it somewhere within the refactoring of ``Sphere`` in the latter part of the topic page.

PR #323, opened 2022-06-10
🎓 Original Sphere Class make a class containing the `Sphere` before `Point3D` 

jameshughes89:
### What

Made a class for the original `Sphere` class

### Why

This version will be the one we introduce the `Sphere` with before refactoring to have a `Point3D` class. I wanted a working version in the repo with unittests for completeness. 

### Testing

Yes, but the simple `assert` tests are defo lacking. That said, I am not worried as this will be pointed out in the topic, saying something like "uhh, feels incomplete, but it's getting mangled..."

### DIFF
--- a/test/test_sphere_original.py
+++ b/test/test_sphere_original.py
@@ -59,10 +59,10 @@ class SphereOriginalTest(unittest.TestCase):
 
     def test_distance_between_centres_various_spheres_returns_correct_distance(self):
         cases = [
-            (Sphere(0, 0, 0, 1), Sphere(Point3D(0, 0, 0), 1)),
-            (Sphere(0, 0, 0, 1), Sphere(Point3D(1, 1, 1), 1)),
-            (Sphere(1, 1, 1, 1), Sphere(Point3D(1, 1, 1), 10)),
-            (Sphere(-1, -1, -1, 1), Sphere(Point3D(1, 1, 1), 10)),
+            (Sphere(0, 0, 0, 1), Sphere(0, 0, 0, 1)),
+            (Sphere(0, 0, 0, 1), Sphere(1, 1, 1, 1)),
+            (Sphere(1, 1, 1, 1), Sphere(1, 1, 1, 10)),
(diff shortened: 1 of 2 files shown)


test/test_sphere_original.py
```diff
@@ -0,0 +1,112 @@
...
+
+from src.sphere_original import Sphere
+
+
+class SphereOriginalTest(unittest.TestCase):
+    def test_sphere_x_returns_correct_x(self):
+        sphere = Sphere(0, 1, 2, 1)
+        self.assertEqual(0, sphere.x)
+
+    def test_sphere_y_returns_correct_x(self):
```
twentylemon, line 11: `correct_y` -- same for `z` test

test/test_sphere_original.py
```diff
@@ -0,0 +1,112 @@
...
+            Sphere(10, 11, 12, 10),
+        ]
+        expecteds = [0, 4.18879, 0, 4188.7902]
+        for (case, expect) in zip(cases, expecteds):
+            with self.subTest(case=case, expect=expect):
+                self.assertAlmostEqual(expect, case.volume(), 5)
+
+    def test_distance_between_centres_various_spheres_returns_correct_distance(self):
+        cases = [
+            (Sphere(0, 0, 0, 1), Sphere(0, 0, 0, 1)),
```
twentylemon, line 61: I'd probably use a [named tuple](https://docs.python.org/3/library/collections.html#collections.namedtuple), or a map for this. Definitely clearer to say `assert case.this.distance(case.that) is whatever` imo.

Though, it's local enough that it's fine as is -- mostly taste.

twentylemon (approved):

test/test_sphere_original.py
jameshughes89: Given the coolness of this idea, I started to change it over, but opted to stop since, although this test class is more for _me_, it is nearly the same as the better `Sphere` class and adding more complexity there (new idea of namedtuple) may be a lot for the students at this stage.

PR #324, opened 2022-06-10
🎓 Models of Computation Topic clean up content

jameshughes89:
### What

Clean up the models of computation topic. 

### Additional Notes

Given the (well needed) additional time spend with objects, I suspect this may be the last topic discussed in the course. Time will tell though as this upcoming Fall semester will be the first run of the course since the significant rework.

### DIFF
--- a/site/topics/models-of-computation/models-of-computation.rst
+++ b/site/topics/models-of-computation/models-of-computation.rst
@@ -176,6 +176,11 @@ Pushdown Automata (PDA) and Context-Free Grammars
     picture of what it's doing exactly.
 
 
+* Although the PDAs are more *powerful* than FSMs, they are still limited in what they can do
+* But the point is, despite their limitations, they can still do quite a bit
+* And the PDAs are, by definition, computers capable of processing information
+
+
 Context-Sensitive Grammars
 ==========================
 
@@ -209,7 +214,12 @@ Context-Sensitive Grammars
     What strings can this system create? Try to generate a few strings and see if you can generalize and see the big
     picture of what it's doing exactly.
 
-    
+
+* Like the CFGs, CSGs have limitations
+* But like before, that does not mean they are not capable of solving/addressing many problems
+* Again, these are *computers*, by definition
+
+
 Turing Machines 
 ===============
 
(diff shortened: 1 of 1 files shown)


jameshughes89:
> I'm not sure the point, really. Throwing "yo dawg, Turing machines are a thing" at them is beyond the scope of the course. What is this priming for?

It's just mentioning it as a bookend, and high-high-level. That said, I totally feel ya'. Any specific ideas?

twentylemon:
Not wholly sure. Anything touched on related to actual computer science will invariably be lost until 3rd/4th year when they actually start doing math. At which point, they'll need to go over it again anyways, since this introduction is surface level.

If the idea here is, "btw, this is actual computer science" -- that's also not helpful I think. Computer science still provides very good foundations for engineering, where I haven't really dealt with turing machines or that type of theory in a long time. (Though, I design grammars fairly regularly, surprisingly.) Point here is, "actual comptuer science" is academic computer science, but real "actual computer science" is both academic and applied, the degree and courses apply to both.

For this filler content, I'd revisit the course objectives, and find some stretch goals in those objectives. I don't think this content would fit those objectives. (Maybe they will -- but an argument for why would be great to see.)

jameshughes89:
> If the idea here is, "btw, this is actual computer science" -- that's also not helpful I think. Computer science still provides very good foundations for engineering, where I haven't really dealt with turing machines or that type of theory in a long time. (Though, I design grammars fairly regularly, surprisingly.) Point here is, "actual comptuer science" is academic computer science, but real "actual computer science" is both academic and applied, the degree and courses apply to both.

I like the idea of emphasizing these ideas. Some CS academics get weirdly pissy about _PrOgRaMmInG iS nOt CoMpUtEr ScIeNcE_ and will misuse that Dijkstra quote as evidence. CS is hilariously broad, and such statements are obnoxious.

twentylemon:
Dijkstra also authored _`goto` considered harmful_, so. Get owned, quote.

jameshughes89:
I ended up hitting Turing Machines a little more, but still very high level. I took it as a chance to impress on them the magnitude of abstraction going on.

twentylemon:
Impress, or frighten?

PR #328, opened 2022-06-13
🎓 Objects Methods Topic `Sphere` assert tests

jameshughes89:
### Related Issues or PRs

Related to #318 

### What

Add the simple `assert` tests for the `Sphere` class. Since `unittest` will be motivated harder in the following topic, I am trying not to poop on the assert tests too hard here, but it is noted that they are getting harder to manage.


twentylemon (approved):
yolo

PR #329, opened 2022-06-13
🎓 Objects Methods Topic `__repr__` method

jameshughes89:
### What

Explain `__repr__` and show how to use it.

### DIFF
--- a/site/topics/objects/methods.rst
+++ b/site/topics/objects/methods.rst
@@ -380,7 +380,7 @@ Magic Methods
     * For example, the ``List`` class' ``__repr__`` returns a string of the form ``["a", "b", "c", "d"]``
 
 * But sometimes, like with a ``Sphere``, it may not be obvious and we just want to get enough information about the ``Sphere`` to be helpful for us
-* If this is the case, a common representation is ``Sphere(1, 2, 3, 4)`` -- class name, and then relevant attribute values within parentheses
+* If this is the case, a common representation is ``Sphere(x=1, y=2, z=3, radius=4)`` -- class name, and then relevant attribute values within parentheses
 
 .. code-block:: python
     :linenos:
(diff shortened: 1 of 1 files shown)


site/topics/objects/methods.rst
```diff
@@ -343,6 +343,72 @@ Magic Methods
...
+* If you were to try this with our newly created ``Sphere`` class, we would get something not overly helpful
+
+.. code-block:: python
+    :linenos:
+
+    sphere = Sphere(1, 2, 3, 4)
+    print(sphere)                   # Results in <__main__.Sphere object at 0x7f99a2edac10>
+
+
+* The default behaviour will be the name of the class along with the memory address of where the object is
```
twentylemon, line 367: Isn't python a VM? That'd be the VM address, not the actual memory address.

site/topics/objects/methods.rst
```diff
@@ -343,6 +343,72 @@ Magic Methods
...
+    * ``print(some_sphere)`` will automatically call it
+    * ``str(some_sphere)`` will call it
+    * ``repr(some_sphere)`` will call it too
+
+* Based on what the object is, there may be a very natural way one would want to represent the object as a string
+
+    * For example, the ``List`` class' ``__repr__`` returns a string of the form ``["a", "b", "c", "d"]``
+
+* But sometimes, like with a ``Sphere``, it may not be obvious and we just want to get enough information about the ``Sphere`` to be helpful for us
+* If this is the case, a common representation is ``Sphere(1, 2, 3, 4)`` -- class name, and then relevant attribute values within parentheses
```
twentylemon, line 383: I typically prefer to name the fields as well. `Sphere(x=1, y=2, z=3, radius=4)` reads a lot clearer.

No need to update the code base imo, but mentioning it in a bullet makes sense -- reinforce that they have control over what they want to see in this case.

twentylemon (approved):

site/topics/objects/methods.rst
jameshughes89: I was hoping the vagueness of that statement, as written, left it such that I didn't need to get into that.

site/topics/objects/methods.rst
jameshughes89: Updated the code anyways --- idk why I was inconsistent in the first place 🤷

PR #336, opened 2022-06-14
🎓 Unittest start `subTest` examples

jameshughes89:
### What

1. Start the `subTest` examples
2. Explain some of the nuance needed for `subTest` 
3. Update unittests to have a float in one of the cases

### Why

Points 1 and 2 should be clear. 
RE 3, this change was made since it used `self.assertAlmostEqual` because `diameter` should work on floats too. Could have changed the test to just be `self.assertEqual` and left the cases alone, but since the functionality is for floats really, I figured it made more sense to go the way I did. 

### Testing

:+1: Re-ran unit tests due to the `radius` change to `10.1` for one of the cases. 

### Additional Notes

Some specific points are emphasized below in comments.

### DIFF
--- a/site/topics/testing/unittest.rst
+++ b/site/topics/testing/unittest.rst
@@ -224,8 +224,7 @@ Subtests
 * In the above example, each test input and expected output were stored in lists
 
     * I used two separate lists, but there is nothing stopping you from using one list of tuples
-
-* The variable names for the lists, ``cases`` and ``expecteds``, were arbitrary and by no means required
+    * The variable names for the lists, ``cases`` and ``expecteds``, were arbitrary and by no means required
 
 * First notice the loop --- there is nothing particularly important for the ``subTest`` here, but the ``zip`` function has not been seen yet
 
--- a/site/topics/testing/unittest.rst
+++ b/site/topics/testing/unittest.rst
@@ -198,7 +198,35 @@ Subtests
 * Although there is nothing wrong with the above tests, we can instead, we can make use of ``subTest`` in this scenario
 
 
+.. code-block:: python
+    :linenos:
+    :emphasize-lines: 15, 16
 
+    import unittest
+
+    class SphereTest(unittest.TestCase):
+
+        # Other test methods not shown for brevity
+
+        def test_diameter_various_spheres_returns_correct_diameter(self):
+            cases = [
+                Sphere(Point3D(0, 0, 0), 0),
+                Sphere(Point3D(0, 0, 0), 1),
+                Sphere(Point3D(1, 1, 1), 0),
+                Sphere(Point3D(10, 11, 12), 10),
+            ]
+            expecteds = [0, 2, 0, 20]
+            for (case, expect) in zip(cases, expecteds):
+                with self.subTest(case=case, expect=expect):
+                    self.assertAlmostEqual(expect, case.diameter(), 5)
+
+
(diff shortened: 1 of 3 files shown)


site/topics/testing/unittest.rst
```diff
@@ -152,6 +152,122 @@ Writing Unit Tests
...
+            :linenos:
+
+            ...
+            for i in range(len(cases)):
+                with self.subTest(case=cases[i], expect=expecteds[i]):
+                    ...
+
+
+* Next, notice ``with self.subTest(case=case, expect=expect)``
+* There is a lot going on here, and much of it we will leave as *important code we need*
```
jameshughes89, line 244: this basically means, *ignore `with`*

site/topics/testing/unittest.rst
```diff
@@ -152,6 +152,122 @@ Writing Unit Tests
...
+
+    * ``self.assertAlmostEqual(expect, case.diameter(), 5)``
+    * This whole portion could be re-written as follows
+
+        .. code-block:: python
+            :linenos:
+
+            ...
+            with self.subTest(abc=case, xyz=expect):
+                self.assertAlmostEqual(abc, xyz.diameter(), 5)
```
jameshughes89, line 262: keyword args explained lightly here. I am hoping the example is enough to get the point across, but again, 👀

site/topics/testing/unittest.rst
```diff
@@ -152,6 +152,122 @@ Writing Unit Tests
...
+    * This just provides an easy way to loop over data within two lists at the same time
+    * This whole portion could be re-written as follows
+
+        .. code-block:: python
+            :linenos:
+
+            ...
+            for i in range(len(cases)):
+                with self.subTest(case=cases[i], expect=expecteds[i]):
+                    ...
```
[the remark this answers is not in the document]
twentylemon: I think replacing the first `...` is doing more damage than good.
```python
cases = [...]
expecteds = [...]
for i in range(len(cases)):
    with self.subTest(case=cases[i], expect=expecteds[i]):
        ...
```

Removing the variable definitions makes it a little confusing at first. Generally though, seems clear enough.

site/topics/testing/unittest.rst
twentylemon: The fact the kwargs have the same names as the variables may be confusing. Clarifying that the left hand sides are the names and right hand sides are the variables may be prudent.

Further, it's kinda not totally true? `subTest` works as a no-args function. The arguments are for documentation only.

jameshughes89: Easiest solution was to remove all the keyword args stuff. This simplifies the whole discussion.

test/test_sphere.py
```diff
@@ -22,7 +22,7 @@ def test_diameter_various_spheres_returns_correct_diameter(self):
         ]
         expecteds = [0, 2, 0, 20.2]
         for (case, expect) in zip(cases, expecteds):
-            with self.subTest(case=case, expect=expect):
```
twentylemon, line 25: May be worth keeping the args around in our own tests though. Distinguishing subtests is useful.

In general, revisiting subtests later to introduce the optional kwargs seems nice to have.

twentylemon (approved):

test/test_sphere.py
jameshughes89: Grand idea bro.

PR #338, opened 2022-06-15
🎓 Course Class add `contains` method

jameshughes89:
### What

Add a `contains` method for the class. Make find start with underscore. 

### Why

1. It felt like it could use one more, idk 
2. Make find "private", or whatever they call it in Python, since I go on a rant in the course content about not accessing the `_students` attribute directly, but then `find` made it seem like, "well, why would I want the index if I am not accessing it the list directly?
3. `_find` is now used as helper for two other methods

### Testing
:+1:

### Additional Notes
There is a magic method, `__contains__` that is more _pythonic_ and uses the `in` operator. I opted to just not do this as I want to avoid more magic methods at this stage. I also never used `in` this way so far in the course; `in` has only been used in loops.

### DIFF
--- a/test/test_course.py
+++ b/test/test_course.py
@@ -48,6 +48,20 @@ class CourseTest(TestCase):
         student_there = Student("Bob", "Smith", 123456789)
         self.assertEqual(0, course.find(student_there))
 
+    def test_contains_singleton_student_does_not_exist_returns_false(self):
+        course = Course("CS101")
+        student = Student("Bob", "Smith", 123456789)
+        course.add(student)
+        student_not_there = Student("Not", "Bob", 987654321)
+        self.assertFalse(course.contains(student_not_there))
+
+    def test_contains_singleton_student_exists_returns_true(self):
+        course = Course("CS101")
+        student = Student("Bob", "Smith", 123456789)
+        course.add(student)
+        student_there = Student("Bob", "Smith", 123456789)
+        self.assertTrue(course.contains(student_there))
+
     def test_remove_singleton_student_does_not_exist_raises_value_error(self):
         course = Course("CS101")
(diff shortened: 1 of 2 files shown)


test/test_course.py
```diff
@@ -100,7 +114,32 @@ def test_find_many_students_duplicate_students_returns_first_occurrence_index(se
         course.add(Student("Niles", "MacDonald", 192837465))
         course.add(Student("Jane", "Doe", 987654321))
         student_there = Student("Jane", "Doe", 987654321)
-        self.assertEqual(1, course.find(student_there))
+        self.assertEqual(1, course._find(student_there))
+
+    def test_contains_many_students_student_does_not_exist_returns_false(self):
```
twentylemon, line 119: Why are these separated from the other `contains` tests?

twentylemon (approved):
Approvals for this content are basically blind at this point. My opinion is that the content shown here is exclusively instilling destructive habits.

jameshughes89:
> Approvals for this content are basically blind at this point. My opinion is that the content shown here is exclusively instilling destructive habits.

Can you be more specific?

twentylemon:
Mutability. Merging data and behaviour. _Object orientation_ is destructive.

jameshughes89:
> Mutability. Merging data and behaviour. _Object orientation_ is destructive.

I feel ya'. Any suggestions on an alternative approach to this based on your experience?

test/test_course.py
twentylemon: I see. I tend group units under test together, instead of the condition under test. That's probably opinion, I find it's a more natural grouping.

twentylemon:
Unfortunately not. At the end of the day, they are concepts that should be taught. The downsides can't really make it into a first year course.

(I nuked two rants.)

PR #342, opened 2022-06-20
🎓 Searching Topic linear search and complexity

jameshughes89:
### What

Show linear search (again) and discuss what _goodness_ is. 

### Why

Given the familiarity with the problem and algo, linear search is a great place to start thinking about the amount of work needed. 

Although the discussion is kept high-level, even with an informal introduction of the ideas around computational complexity, they can start to get a sense of what _good_ is. 

The ideas will not be on the test, but it will benefit the keen students as they will have a framework to start thinking more about their algos. 

### Where

I stopped before the activity of binary search. That will become a new section in a subsequent PR. 

### How

Complexity is kept high level. I don't want to over do it as it can feel intimidating if we get formal. 

### Additional Notes

1. I am trying to keep it informal and intuitive, but let me know if I over did it one way or the other. 
2. I was thinking of ending the section with something about "we need something to compare to now though", but opted to not do that here since it may fit better after binary search as it will hopefully be very obvious at that point.


twentylemon (approved):
Given it's not tested and serving as an intro to second semester, the depth seems fair to me.

PR #343, opened 2022-06-20
🎓 Searching Topic binary search

jameshughes89:
### What

Discuss/show binary search and analyze it at a high level. 

### Why

1. Binary search is cool
2. Binary search shows a clear example of how the structure of the data can help
3. It also has the neat recursiveness --- half of half of half of half...
4. It gives us something to compare linear search to
5. It is a nice segue to sorting, I guess

### How

Again, tried to keep it high level. In fact, there is no code for binary search as it is intended to be kept to the level of intuition.

### Additional Notes

There is a lot more math-ie stuff going on in this analysis, but I *think* it's still high level and informal enough to be accessible (tell me if you disagree, but remember).


site/topics/searching/searching.rst
```diff
@@ -54,32 +54,77 @@ Searching (again)
...
+=============
+
 .. admonition:: Activity
     :class: activity
 
-    James will now guess a number between 0 and 1023 in less than 10 guesses. 
-   
-.. admonition:: Activity++
-    :class: activity
+    I will guess a number between :math:`0 -- 1023` in :math:`10` or less guesses.
```
twentylemon, line 64: May be more intuitive to search for a word in a dictionary. People use interpolation search (looking for "poop" you'd open the book to where you'd think P is, then correct based on your understanding of the distribution of first letters of words), binary search is basically that except you ignore any statistical properties of the underlying data -- close enough.

...are zoomers aware of dictionaries? maybe the number thing is fine

site/topics/searching/searching.rst
```diff
@@ -54,32 +54,77 @@ Searching (again)
...
+    * If there are :math:`n` things in the ``haystack``, I have to look at all :math:`n`
+    * If we doubled the size to :math:`2n`, the amount of things needed to be looked at also doubles to :math:`2n`
+
+* With binary search however, the relationship between the size of the input :math:`n` and the amount of work is :math:`log_{2}(n)`
+
+    * Doubling the size to :math:`2n` only adds one more guess
+
+        * :math:`log_{2}(2n)`
+        * :math:`log_{2}(n) + log_{2}(2)`
+        * :math:`log_{2}(n) + 1`
```
twentylemon, line 118: This is probably unnecessary. Intuitively, doubling the size adds only one step to the progression above that starts at 1024.

twentylemon (approved):

site/topics/searching/searching.rst
jameshughes89: Agreed. Additionally, if anything, including it is worse:
- They see more logs and freak our
- They think the logs are important
- They don't see the forest through the trees

site/topics/searching/searching.rst
jameshughes89: Here's why I'll leave it:
- Interpolation search is how it is introduced in the subsequent course
- _Interpolation search -> binary search -> analysis_ is more dots to connect than _binary search -> analysis_
- They probably have played this higher/lower game before anyways
- And above all else, which cannot be stressed enough, like at least 99.98% of the reason, that change would require more than 0 effort.

twentylemon: I wasn't suggesting going into interpolation search, rather that people intrinsically use that algorithm. The gap from interpolation to binary is trivial, so an example where people use interpolation search inherently may be a better example.

Numbers are still fine.

PR #345, opened 2022-06-21
:mortar_board: Sorting Topic introduce the topic

jameshughes89:
### What

Introduce and motivate the sorting topic.


jameshughes89:
I hit a button --- re-opened.

PR #346, opened 2022-06-23
🎓 Sorting Topic selection sort

jameshughes89:
### What

Discuss selection sort. Also include code for selection sort. 

### How

When getting into specifics, I made reference to a _list_ as opposed to some general collection. This was because the list are clear concrete things at this stage that they are famillar with.

### Testing

:+1:

### Additional Notes

I was originally going to keep the code out of it, but ultimately added it.


site/topics/sorting/sorting.rst
```diff
@@ -38,7 +38,78 @@ Sorting Algorithms
...
+
+
+* Selection sort is fairly accessible, and chances are you have sorted things in real life using this algorithm
+* Selection sort works by repeatedly selecting the smallest element from the collection
+* The high-level algorithm is as follows
+
+    * Start with the unsorted list and an empty list for the sorted elements
+    * For each element in the collection
+
+        * Perform a linear on the unsorted list for the current smallest element
```
twentylemon, line 58: linear _search_

site/topics/sorting/sorting.rst
```diff
@@ -38,7 +38,78 @@ Sorting Algorithms
     high-level idea, not the actual, literal implementations. This idea will be emphasized as we discuss the individual
     sorting algorithms below.
 
-   
+
+Selection Sort
```
twentylemon, line 42: I always found this algorithm to be kinda a waste of time. The algorithm is basically identical to insertion sort, just getting the "next" element instead of placing the "current" element in the right order. Plus, it's worse in basically all regards to insertion sort in practice (unless your memory is spatially distant or like, a tape drive).

I'm not sure if bubble sort is any more intuitive, but it's quite a bit different from other sorts.

test/test_sorting_topic.py
```diff
@@ -0,0 +1,33 @@
+import unittest
+
+from src.sorting_topic import selection_sort
+
+
+class SortingTopicTest(unittest.TestCase):
```
twentylemon, line 6: Given you're cramming everything into a single file here, this test class is going to get chonky. Consider pulling out the list of cases into a class constant and using them all with `subTest`. Every sorting algorithm you add is going to be another ~30 lines of the same tests, otherwise.

site/topics/sorting/sorting.rst
jameshughes89: Early on, it's more about having easily identifiable things they can recognize as having done before.

site/topics/sorting/sorting.rst
jameshughes89: rip

site/topics/sorting/sorting.rst
twentylemon: I feel like I do insertion sort when sorting a shelf of books or something.

jameshughes89: It's always interesting how the sorting first plays out:
- Insertion and Selection are all ones they agree they have done before
- Bubble sort is ones they may or may not feel like they've done before, but nearly always, when asking the class for a sorting algo for the first time, they try to explain something nearly bubble sort. The usual hangup is the repeated scans.

jameshughes89: Except that one kid that apparently uses quicksort to sort all his stuff 🙄

twentylemon: I use bogosort to cheat at cards.

jameshughes89: _pog_-o sort

twentylemon: But, yeah. I'm not necessarily advocating bubble sort to start with. Insertion sort is more intuitive than that, and it's much more useful than selection sort (most standard library sorting algorithms are quicksort when array large, then use insertion sort when array small). Given the similarity to selection sort, I'm moreso saying just skip selection sort. Maybe show bubble sort as "yo, there's other comparison sorting algorithms" that don't use recursion.

101 doesn't _really_ go into sorting anyways. Going a little light makes sense to me.

jameshughes89: Idk ---- I like the possible progression of selection -> insertion 

1. Build a new list by always tacking on the next proper element to the end 
2. Now, instead, just add stuff to the new list by putting the things in their proper place within it 
3. ...Bubble is weird. 

I suppose the takeaway is:
- Think about a problem you all have experienced 
- See different algos for the same problem
- Think a bit about the algos

jameshughes89: I spent the day thinking about it and I still think including Selection Sort is worth while at this stage. I took your perspective and challenged the fact that _this is how I did it every other year_, but I still think it's worth showing the 3 given the big part of identifying a strategy they have definitely done in real life.

**However** , with that said, my current though process is to either (a) leave it as 3 distinct sections, or (b) have a section on Insertion Sort, but then have a subsection (or not even, perhaps just a note or something) on Selection Sort within the section. 

I'll let the ideas brew a little.

twentylemon: All right. If you're convinced, I won't push. Questioning the status quo is good enough.

twentylemon (approved):

jameshughes89:
I'm gonna smush it, but acknowledge that it may be altered after other sorts get added.

PR #347, opened 2022-06-24
🎓 Sorting Topic add insertion, bubble, and another bubble

jameshughes89:
### What

Add `insertion_sort`, `bubble_sort`, and `bubble_sort_improved` (bubble sort, but uses a flag to check if any swaps took place) to the repo. 

### Why

They will be discussed (though, the improved bubble will probably just be a footnote in the topic). 

### Testing
:+1:

### DIFF
--- a/src/sorting_topic.py
+++ b/src/sorting_topic.py
@@ -8,3 +8,35 @@ def selection_sort(collection):
         collection.remove(current_smallest)
         sorted_collection.append(current_smallest)
     return sorted_collection
+
+
+def insertion_sort(collection):
+    sorted_collection = []
+    for element in collection:
+        i = 0
+        # Scan sorted collection to find insertion spot
+        while i < len(sorted_collection) and sorted_collection[i] < element:
+            i += 1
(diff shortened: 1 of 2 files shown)


src/sorting_topic.py
```diff
@@ -8,3 +8,35 @@ def selection_sort(collection):
         collection.remove(current_smallest)
         sorted_collection.append(current_smallest)
     return sorted_collection
+
+
+def insertion_sort(collection):
+    sorted_collection = []
```
twentylemon, line 14: Insertion sort is typically in-place. There's no need to make a copy of the entire collection.

Really, same goes for all sorts. Either sort in-place or return a sorted copy while leaving the original untouched.

jameshughes89: _Not_ in place was intentional as I feel having a new list will be more accessible at this stage. Bubble sort does not really lend itself to not in-place, but good call on having it return a copy to have no side effect.

twentylemon (approved):

PR #361, opened 2022-08-29
📝 Bump flake8 isort

jameshughes89:
### What

Update version of flake8-isort to 4.2. 

### Why

fix the one thing.


jameshughes89:
close because in like 2 days the robot will fix it

PR #364, opened 2022-08-30
🆕 Syllabus --- Assignment Recoupment

jameshughes89:
### What

Explain _assignment recoupment_. 

### Why

This is something I did back in like 2016/2017. The idea is, students can take their returned assignments, correct mistakes, and recoup 50% of the lost marks. 

Eg. 
- A student gets 80/100
- They correctly fix things that account for 15 of the 20 lost marks
- They get +7.5 on the assignment for a total of 87.5/100. 

Motivation:
(a) Students are expected to make mistakes and are an essential part of learning
(b) Marks are not intended to be a reward/penalty; marks are for feedback
(c) We want students to actually look at their assignment feedback and use it as a learning opportunity 
(d) Marks tend to be very bimodal, and struggling students stand to benefit the most from this idea
(e) Students that actually care are the ones that will do this
(f) The marker will be using unit tests, so they will have more hours to dedicate to this
(g) It worked well before

The reason for the limitations on eligibility is to prevent students from submitting nonsense and using this as a mechanism to hand in late work. Basically, students can't vomit on the page and expect to get feedback to resubmit. 

### Additional Notes

The potential for ambiguity in the explanation is ripe for abuse and complaints, but fortunately the times I did this in the past for much larger courses, it ended up not being a problem --- though, those were different times... I considered adding a grade cutoff for eligibility, like, only submissions that get above 60% can resubmit, but I can see that getting hairy. 

Overall the TA will err on the side of _being nice_ in terms of eligibility, but at the end of the day, the constraints are in the syllabus so there is something to fall back on.

### DIFF
--- a/site/outline/outline.rst
+++ b/site/outline/outline.rst
@@ -224,6 +224,10 @@ only assignments deemed eligible may be resubmitted. Only assignments that were 
 eligible. Students with missing code/functions/portions of the assignment will not be eligible. If the student has
 portions of code that demonstrate a lack of a sincere attempt, the assignment will not be eligible for recoupment.
 
+Further, it is the responsibility of the student to make clear to the marker what exactly has been updated and changed.
+If the marker is unable to quickly determine what fixes the student has made, the resubmission will not be considered
+for recoupment. This could be in the form of notes in the Moodle submission with corresponding comments within the code. 
+
 
 Tests
 =====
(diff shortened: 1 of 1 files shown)


twentylemon (approved):
I hope your TA's get paid more because of this.

PR #370, opened 2022-09-02
💣 🎓 Recursion --- Eliminate Topic and Lab

jameshughes89:
### Related Issues or PRs

Related-ish to #369 

### What

Remove the topic and lab on recursion. 

### Why

Although this is maybe my favorite topic to get into in lecture, it is not likely that we will get this far in the course. Further, it always kinda' fell flat on the students. By this time in the semester, it's a lot to chew on for them. 

### Additional Notes

**If** we happen to get this far, we can add it back in.


jdelamer (approved):

twentylemon (approved):
You could consider keeping the lecture notes and lab files, and only remove the topic from the main lists. This would serve as a reminder of stuff you did have, and make it 2% easier to re-add the content (`git revert` takes more time). Could also add notes (ie, comments in the file) about why it was removed.

That said, searching commit history for :bomb: is also fine, and commits link to pull requests which include context. Just a thought.

jameshughes89:
> You could consider keeping the lecture notes and lab files, and only remove the topic from the main lists. This would serve as a reminder of stuff you did have, and make it 2% easier to re-add the content (`git revert` takes more time). Could also add notes (ie, comments in the file) about why it was removed.
> 
> That said, searching commit history for :bomb: is also fine, and commits link to pull requests which include context. Just a thought. 

Though great suggestions, even still, after reflection, the inclusion of both recursion and models of computation was not great. I love talking about them, but that's really the only reason why I had them in the content. Although important, and there is an argument for including them, there's a stronger argument for spending more time on some of the earlier content for better prep going forward.

PR #381, opened 2022-09-06
:new: Values --- Add SSD

jameshughes89:
### What

Add SSD to the part where HDDs are mentioned

### Why

They are a thing that are common now


twentylemon (approved):
maybe better to use the umbrella term, storage

PR #410, opened 2022-09-16
:bug: Functions Topic --- Fix bad math :microscope: 

jameshughes89:
### What

Change the math in an example to be correct. 

### Why

The square of the sum of 3 & 3 is not 81 --- it's 36. 

### How

I did the math correctly this time. 

### Testing

I used a calc to make sure I was not losing it. 

### Additional Notes

Kinda' urgent.


twentylemon (approved):
3 + 3 is 9. Close enough.

jdelamer (approved):

PR #414, opened 2022-09-20
:mortar_board: Assignment 2 --- Present task and discuss provided file

jameshughes89:
### What

Introduce the task they are to do for Asn2. Show the provided file. Also change the warning to a note since it's more of a note than a warning. 

### Additional Notes

The actual provided files are not changed yet; they are the old versions. New versions will be in another PR.

### DIFF
--- a/site/assignments/zombies/zombies.rst
+++ b/site/assignments/zombies/zombies.rst
@@ -36,6 +36,21 @@ You will
 Provided Files
 ==============
 
+You are provided with
+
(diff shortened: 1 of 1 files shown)


twentylemon (approved):

twentylemon:
I do think this assignment should kinda change to be Conway's game of life. `matplotlib` supports animation, so you could provide the visualization aspect of it in theory.

jameshughes89:
Not a bad idea at all. 

Ideally, going forward, there will be a suite of assignments for each of the 4 assignments.

Not to mention the fun thought experiments one could get into with game of life.

PR #438, opened 2022-10-04
:mortar_board: Assignment 2 --- Using the code for analysis

jameshughes89:
### What

Explain how to run the code to do a quick analysis of the results of their simulations. 

### Why

I think it's fun. 

### Additional Notes

I am worried about the complexity of generating a distro of results, so I am considering giving them the code the running the simulation 100x. What y'all think?


twentylemon (approved):
This doesn't seem like it's too difficult. If they get the whole simulation loop part, this is essentially the same idea. Even simpler though, since you don't need to keep track of a before state.  
Where I may be worried is the overall assignment length. I don't really have bearings on this, but simulation step + simulation + graph + write up sounds upper year (though, the write up I imagine is a few sentences).

PR #457, opened 2022-10-20
:lipstick: Exceptions Topic --- Cleanup last section :microscope: 

jameshughes89:
### What

Clean up last section of topic

### Why

- There was bad typesetting 
- Added "ultimately" to make the point flow from the previous better
- Time should not have a capital, and i made it italics

### DIFF
--- a/site/topics/exceptions/exceptions.rst
+++ b/site/topics/exceptions/exceptions.rst
@@ -211,9 +211,9 @@ Why Care?
 * The trouble is, that depends on your situation
 * The point is, how can the programmer in 1991 know what you want to do with your situation today?
 
-# Exceptions are fantastic because
+* Ultimately, exceptions are fantastic because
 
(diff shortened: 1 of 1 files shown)


site/topics/exceptions/exceptions.rst
```diff
@@ -211,9 +211,9 @@ Why Care?
 * The trouble is, that depends on your situation
 * The point is, how can the programmer in 1991 know what you want to do with your situation today?
 
-# Exceptions are fantastic because
+* Ultimately, exceptions are fantastic because
```
twentylemon, line 214: _fantastic_ is probably a little strongly worded. It is arguable that exceptions are side effects, a lot of people like to use constructs like `Either` or `Try`, similar to python's `Option` for nullability.

twentylemon (approved):

PR #465, opened 2022-10-24
:mortar_board: Tic-Tac-Toe Assignment --- `is_move_valid` function description

jameshughes89:
### What

- Describe what makes a move valid
- Discuss a function to determine if a move is valid --- `is_move_valid`
- Provide examples of using the function
- Warn about indexing order

### DIFF
--- a/site/assignments/tic-tac-toe/tic-tac-toe.rst
+++ b/site/assignments/tic-tac-toe/tic-tac-toe.rst
@@ -108,6 +108,10 @@ Part 3 --- Parse Move
 Part 4 --- Validate Move
 ========================
 
+Player moves are considered valid if (a) the specified game board cell/location is unoccupied (contains a ``" "``
+(space) character) and (b) is within the game board.
+
+
 
 Part 5 --- Apply Move
 =====================
--- a/site/assignments/tic-tac-toe/tic-tac-toe.rst
+++ b/site/assignments/tic-tac-toe/tic-tac-toe.rst
@@ -120,9 +120,15 @@ For example, consider the current game board being ``board = [["X", " ", " "], [
 * ``is_move_valid((2, 1), board)`` returns ``False`` since ``(2, 1)`` already contains an ``"O"``
 * ``is_move_valid((-2, 1), board)``returns ``False`` since the move location does not exist on the provided game board
 
+
 .. note::
 
-    When thinking of the game board like a matrix
+    When thinking of the game board like a matrix, one may be tempted to think of indexing the board with ``x`` first
+    and ``y`` second --- ``board[x][y]``. However, the convention being followed is that ``x``, like in a cartesian
+    coordinate system, specifies the horizontal positioning --- the column --- and ``y`` specifies the vertical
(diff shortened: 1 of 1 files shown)


site/assignments/tic-tac-toe/tic-tac-toe.rst
```diff
@@ -108,6 +108,28 @@ Part 3 --- Parse Move
...
+
+
+.. note::
+
+    When thinking of the game board like a matrix, one may be tempted to think of indexing the board with ``x`` first
+    and ``y`` second --- ``board[x][y]``. However, the convention being followed is that ``x``, like in a cartesian
+    coordinate system, specifies the horizontal positioning --- the column --- and ``y`` specifies the vertical
+    positioning --- the row. This would mean that one needs to index the board with ``y`` first to specify the row and
+    then once the row is selected, the ``x`` value is used to indicate which column in the row the cell/location is. In
+    other worse, the correct indexing would be ``board[y][x]``.
```
twentylemon, line 131: This seems implementation specific. It onlys factor in if you specifically write `render` as nested foreach loops, without indexes. However, if you use `for (x,y) in (0..n, 0..m)`, then it does not seem like it'd apply.

The explanation here also leaves a little to be desired. The conclusion does not follow the premise. "We want cartesian like coordinates" -> "index like `board[y][x]`" simply does not follow. Needed is the fact that one may consider `board[i]` to be the _ith_ row of the board, ie the row with the y value of _i_ -- but as mentioned, you could also consider it to be the _ith_ column.

jameshughes89: > This seems implementation specific. It onlys factor in if you specifically write render as nested foreach loops, without indexes. However, if you use for (x,y) in (0..n, 0..m), then it does not seem like it'd apply.

I'll work on the wording --- the intent is to set the constraint to help eliminate the potential issues they run into with this. 

Interesting note, although in no way limiting, `numpy` has the default printing convention to have it be [row][col]. 

Screenshot from 2022-10-25 09-36-01

> The explanation here also leaves a little to be desired. The conclusion does not follow the premise. "We want cartesian like coordinates" -> "index like board[y][x]" simply does not follow. Needed is the fact that one may consider board[i] to be the ith row of the board, ie the row with the y value of i -- but as mentioned, you could also consider it to be the ith column.

This was discussed in the setup game section in #462. That said, it is defo still worth consistent clarity here.

jameshughes89:
Updated. 

I was considering adding a figure showing the difference in thinking of the first index as a row vs. col, but (a) it may end up just being info overload, and (b) it requires marginally more work than not.

twentylemon (approved):

site/assignments/tic-tac-toe/tic-tac-toe.rst
twentylemon: > Interesting note, although in no way limiting, numpy has the default printing convention to have it be [row][col].

This is how I've typically done this kind of thing in the past too, it is a little more natural. It's more simpler to display a list horizontally than it is to display it vertically, you can do foreach loops on the board (like `board.foreach(println)` and you're basically done), but you can still also loop over indexes.

PR #466, opened 2022-10-24
:mortar_board: Tic-Tac-Toe Assignment --- `apply_move`

jameshughes89:
### What

Describe how to apply the move. Take care to warn about shallow vs. deep copy. 

### Testing

~I'm going to check the artifact to see if it worked.~
Looks good.


site/assignments/tic-tac-toe/tic-tac-toe.rst
```diff
@@ -112,6 +112,30 @@ Part 4 --- Validate Move
...
+lists and not just the outside list. If we perform a copy naively, we may accidentally make a copy of the list with
+copies of the references --- this is called a "shallow copy". Refer to the following images to see the difference
+between a "shallow" copy and a "deep" copy in this context.
+
+.. figure:: shallow_copy.png
+
+    Example of a "shallow copy" --- only the references to the internal lists were copied. The actual internal lists
+    were never copied.
+
+.. figure:: deep_copy.png
```
twentylemon, line 135: The deep copy figure is a little excessive with the copying. Mutations to immutable collections tend to re-use as much as possible from the original -- they structurally share. The trivial example is appending to a list, which re-uses the entire source list (the new element's `next` is the source list).

In this case, the only element of the board that actually needs to be copied is the one the move is actually being applied to. The others are still "immutable," so you can reference them freely.

Such an approach is _probably_ too much to go into for first year, but maybe worth a small note for the keeners?

twentylemon (approved):

site/assignments/tic-tac-toe/tic-tac-toe.rst
jameshughes89: funnily enough, we went over this exactly in class today as an aside.

PR #478, opened 2022-10-27
:mortar_board: Tic-Tac-Toe Assignment --- Introduce render and `render_cell`

jameshughes89:
### What

1. Motivate render functions
2. Explain single `render_cell` function

### Additional Notes

`render_row` and `render_board` will be in subsequent PRs.

### DIFF
--- a/site/assignments/tic-tac-toe/tic-tac-toe.rst
+++ b/site/assignments/tic-tac-toe/tic-tac-toe.rst
@@ -263,7 +263,15 @@ human friendly string that can be displayed. This functionality will be broken d
 Render Cell
 -----------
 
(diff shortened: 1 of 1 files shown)


twentylemon (approved):

twentylemon:
> `render_row` and `render_board` will be in subsequent PRs.

Imo, these are backwards. It may be more instructional to start at the end (`render_board`) and introduce new pieces as we find the problem to be too large. `render_board` is chonky, so let's loop over `render_row` and add filler rows. `render_row` is kinda chonk as well, so let's have `render_cell` and filler columns.

jameshughes89:
For simplicity, I will merge the approved relevant PRs, but massage the order with a new PR.

PR #481, opened 2022-10-27
:bug: 2D List Lab 8 --- Indent images to fix numbering

craiglandry16:
### What

changed the before kattis question numbers so they are in correct order by indenting images. 

### Why

idk, felt like it lol

### Where

lab 8-2D lists

### How

on here, idk if i did it right


jameshughes89 (approved):
Awesome

PR #507, opened 2022-11-07
:mortar_board: Country Catalogue Assignment --- Explain `Country` class

jameshughes89:
### What

Describe expectations for the `Country` class.


site/assignments/country-catalogue/country-catalogue.rst
```diff
@@ -27,22 +27,44 @@ Part 1 --- Uploading Files to Colab
...
 -----------------------------
 
+Write a method ``population_density`` that returns the population density of the ``Country`` object. The population
+density should be a float.
+
 
 Part 4 --- Equals and Repr
 --------------------------
 
+Write the ``__eq__`` magic dunder method for the ``Country`` class. For our purposes, two ``Country`` objects will be
```
twentylemon, line 54: It's _magic method_ or _dunder_. Prefer to use _magic method_ since that is what the python documentation typically refers to them as, though they are officially called [special methods](https://docs.python.org/3/reference/datamodel.html#special-method-names).

twentylemon (approved):

site/assignments/country-catalogue/country-catalogue.rst
jameshughes89: I intentionally went with "magic dunder" to make sure I piss everyone off. 

Same reason I say Gee Eye Eff for gif. It just pisses everyone off.

PR #541, opened 2022-12-02
:robot: Python version --- Change to 3.11

jameshughes89:
### What

Update to python 3.11

### Testing

:+1:

### Additional Notes
Idk if I really wanna' merge yet. However, I do want to see if the actions work.

### DIFF
--- a/.github/workflows/sphinx-build.yml
+++ b/.github/workflows/sphinx-build.yml
@@ -17,7 +17,7 @@ jobs:
     - name: Set up Python
       uses: actions/setup-python@v4
       with:
-        python-version: 3.10.0
+        python-version: 3.11.0
     - name: Install Dependencies
       run: |
         python -m pip install --upgrade pip setuptools wheel
@@ -36,7 +36,7 @@ jobs:
     - name: Set up Python
       uses: actions/setup-python@v4
       with:
-        python-version: 3.10.0
+        python-version: 3.11.0
     - name: Install Dependencies
       run: |
         python -m pip install --upgrade pip setuptools wheel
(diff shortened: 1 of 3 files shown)


.github/workflows/sphinx-build.yml
```diff
@@ -17,7 +17,7 @@ jobs:
     - name: Set up Python
       uses: actions/setup-python@v4
       with:
-        python-version: 3.10.0
```
twentylemon, line 20: Any reason to force specify `3.11.0`? `3.11` seems specific enough.

jameshughes89: Legit, because it was 3.10.0, and before it was 3.9.0. 

From the docs, having only 3.x would be "faster".

jameshughes89: I used bad syntax there. 

I meant 3.SOMETHING without the 3rd part of the version number.

twentylemon: Gotcha. That is how duckbot specifies [3.8](https://github.com/duck-dynasty/duckbot/blob/main/.github/actions/setup-python-venv/action.yml#L26). It is the way to go, imo.

jameshughes89:
It's just 3.11 now. 

I'll make sure it's happy on windows before I merge. Other repos will be updated once this one is all good.

jameshughes89:
Removed the legacy thing for the windows venv setup description since it's not needed. 

Everything was fine on windows after updating locally. PyCharm seems to be convinced that it's still using 3.10 though, but it's defo not... 🤷 

image
image
image
image

jameshughes89:
> Removed the legacy thing for the windows venv setup description since it's not needed. 
> 
> Everything was fine on windows after updating locally. PyCharm seems to be convinced that it's still using 3.10 though, but it's defo not... 🤷 
> 
> image
> image
> image
> image
> 

I found people talking about a similar issue with pycharm showing the wrong python version online. Some said it resolved after a pycharm update. I'll see what happens.

twentylemon (approved):

PR #542, opened 2022-12-21
💣🎓 Assignment FAQ --- Remove potential confusing wording

jameshughes89:
### What

Remove all wording from the "Can I work with a friend?" question that is not "no".

### Why

The academic integrity committee felt that:

> After several hours of reading your case files, conducting interviews with your students, and cross-checking, the committee found there were inconsistencies within your course syllabus and FAQ dealing with what constitutes a collaboration - when this is appropriate, at what "level" this should be done, by whom, if, when and how this should be documented. Due to the ambiguities leading to student confusion on this matter, the committee settled on its decision as presented. Further, the committee strongly recommends you revisit your course syllabus and FAQ to alleviate these issues and impact future cases.

Having gone through what they referenced, I cannot find any ambiguity from within the syllabus as the wording is copy/pasted from the university's academic policies. When looking at the FAQ, this was the only area of potential concern I could find. 

### How

Only keep the word "no". 

### Additional Notes

1. I feel that it is unfortunate since I did like the wording as is. Further, I do not feel that the wording, especially within the context of the course and how it was conducted, to be ambiguous or inconsistent. Additionally, colleagues agreed that it was not ambiguous. To be fair, however, it is possible that this is not what the committee was referring to. 
2. Given how the committee has operated and the direction the committee has taken over the past few years, I suspect that, realistically, this was a _post hoc_ explanation from the committee to justify their decision.


jameshughes89:
I added some emphasis on independent learning to help make clear that _Googling X_ is to facilitate independent learning.

jdelamer:
I think there was a mistake.

jameshughes89:
> I think there was a mistake.

Good catch

twentylemon (approved):
Does that committee have a document you can link to instead of trying to replicate their guidelines?

jameshughes89:
> Does that committee have a document you can link to instead of trying to replicate their guidelines?

The trouble is, they do, but they don't follow their guidelines so if anything that would make things more confusing.

twentylemon:
Still seems better to link to them. Them improving is better for everyone, instead of only your classes.

Issue #195, opened 2022-05-02
💄 Change Activity Colours


jdelamer:
What
Update/change the colours for the activity scheme.

Why
The dark blue colour is not 'nice'.

How
Modifying custom.css

jameshughes89:
Closing this because the activity colours appear to have already been updated in the current site styling. The custom CSS now defines a distinct activity colour scheme, so this no longer looks like an active issue.

PR #552, opened 2023-01-16
 :lipstick:  Update css and create other admonition classes

jdelamer:
:lipstick: 

### Related Issues or PRs

Following #195

### What

Update the custom.css file with new admonition classes and colors.
Create a JS file to give a "mac" header to code-block and a clipboard button to copy the code.

### Why

Improve the color of the activities and other admonitions.

### Where

Everywhere.

### How

New CSS and JS files and modification of conf.py.

### Testing

image

### Additional Notes

How to use the new admonition classes:

- Activity:

```
.. admonition:: Activity
   :class: activity
```
- Example:

```
.. admonition:: Example
   :class: example
```

### DIFF
--- a/site/_static/css/custom.css
+++ b/site/_static/css/custom.css
@@ -77,10 +77,10 @@ div.highlight div.mac-header span.dot {
   width: 14px;
   height: 14px;
   border-radius: 50%;
-  background: rgba(0,0,0,0.1);
+/*   background: rgba(0,0,0,0.1); */
   margin: 0 4px 0 0;
 }
-div.highlight:hover div.mac-header span.dot.red {
+/* div.highlight:hover div.mac-header span.dot.red {
   background: rgb(255,86,80);
 }
 div.highlight:hover div.mac-header span.dot.yellow {
@@ -88,7 +88,7 @@ div.highlight:hover div.mac-header span.dot.yellow {
 }
 div.highlight:hover div.mac-header span.dot.green {
   background: rgb(41,193,68);
-}
+} */
 div.highlight div.mac-header span.copy {
   position: absolute;
   top: 0;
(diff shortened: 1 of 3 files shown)


jameshughes89:
[I checked the build artifact](https://github.com/jameshughes89/cs101/actions/runs/3934401155) and it seems that the changes did not actually make a change. See the below image. Can you build locally with sphinx to check?

image

jameshughes89:
I really like the colours, but am questioning the mac style thing for the code. 

1. Is there specific motivation for the mac style?
2. If there is not any particular motivation, is there a way to just remove the 3 circles in the top left?

jdelamer:
Regarding the build.
When sphinx build the website it doesn't "update" the _static folder, so the new css and js files are not copied.
I don't know if there is a command to force it.

For the mac style, there is not a specific reason. I just find it neat, but I can remove the three circles.

jameshughes89:
> Regarding the build. When sphinx build the website it doesn't "update" the _static folder, so the new css and js files are not copied. I don't know if there is a command to force it.

Hmm... let's look into this. 

> For the mac style, there is not a specific reason. I just find it neat, but I can remove the three circles.

Let's see how it looks without the 3 circles and go from there.

jdelamer:
Without the circles it gives the following:
image

jameshughes89:
> Without the circles it gives the following: image

Nice. Is the copy button in the top left functional?

jdelamer:
Yes it is.

jdelamer:
I think the commend is ``-a`` to write every files each time.

jameshughes89:
Just checked the artifact --- looks good. 
image
image

jameshughes89 (approved):

PR #567, opened 2023-08-15
:mortar_board: Assignments --- Remove due dates and weights :microscope: 

jameshughes89:
### What

1. Remove the weights of the assignments
2. Remove the due dates

### Why

1. Idk yet if I will make them 5, or 10, or some 5 and some 10. 
2. They are TBD

### Testing

No way jose

### DIFF
--- a/site/assignments/country-catalogue/country-catalogue.rst
+++ b/site/assignments/country-catalogue/country-catalogue.rst
@@ -2,8 +2,8 @@
 Country Catalogue
 *****************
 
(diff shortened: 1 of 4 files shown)


twentylemon (approved):
Change all this info (#566 as well) to be variables or something may make sense. Then at least you update one spot and it affects everywhere, and you don't forget places.

At the same time, it's for something that occurs 1-3x a year, so maybe /shrug

PR #578, opened 2023-09-08
💄 Topic variables types --- Remove word 🔬 

jameshughes89:
### What

Remove the 2nd use of the word "both" 

### Why

It was weird with the 2 of them


twentylemon (approved):
both both

PR #582, opened 2023-09-20
:new: :mortar_board: Lab testing --- Make new lab on testing 

jameshughes89:
### What

Make new lab on testing, type hints, and docstrings 

### Why

Was not far enough along in lecture topics to hit the conditionals lab, which is currently the 3rd lab 

### Testing

:+1: 

### Additional Notes

1. Some of the kattis problems I added are trivial. They wanted more though, so at least they have the skills to do these ones. 
2. More detail _could_ have been provided on the 3 "main" lab questions. For example, I didn't explain what \Delta v means. This was intentional as anything they don't know they should be able to look up, thus it is an exercise in _figuring it out_. 
3. The last question is taken from assignment 1.

### DIFF
--- a/site/index.rst
+++ b/site/index.rst
@@ -97,6 +97,7 @@ Lab Sections
    
     labs/hello-world/hello-world
     labs/functions/functions
+    labs/testing/testing
     labs/conditionals/conditionals
     labs/car-rental/car-rental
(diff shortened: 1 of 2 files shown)


twentylemon (approved):
Seems sane

PR #604, opened 2023-11-30
:technologist: :bug: Country Catalogue Assignment --- Remove mention of `area` in `add` docstring

jameshughes89:
### What

Remove the mention of the `area` parameter in the `add` method in the `CountryCatalogue` class. 

### Why

It's not actually a parameter for the method. 

### Testing

YOLO

### Additional Notes

1. I believe this is from a previous version of the assignment where the `add` method created the `Country` object and then appended. 
2. Oddly, this issue did not exist in the assignments repo.


site/assignments/country-catalogue/asn4.ipynb
jameshughes89, line 1: These changes are BS. It's only the docstring.

jameshughes89:
I changed out the ipynb file by downloading from colab to eliminate any issues with the pycharm editing of the ipynb file. 

Also, looks like the ipynb file keeps track of how many times a cell was run, which is neat. Those fields in the file are changed here but should not have any functional implication.

twentylemon (approved):

PR #621, opened 2024-05-14
:mortar_board: Course Schedule --- Create a schedule for the async version of the course

jameshughes89:
### What

Create a course schedule for the online asynchronous version of the course

### Why

Given that the course is asynch, they need something to give them a schedule. 

### Where

Put it under the outline section

### How

1. I figured I would have an assignment due every 2 weeks
2. I looked at what topics would need to be known for each assignment
3. I then also took into consideration the amount of content is in each of the topics
4. I allllso took into account the topics that would _make sense_ to be grouped together

### Testing

Built local, looks good. 

### Additional Notes

The specific dates of assignments and tests may change as I do not know all the details of online summer courses here

### DIFF
--- a/site/outline/schedule.rst
+++ b/site/outline/schedule.rst
@@ -1,0 +1,84 @@
+===============
+Course Schedule
+===============
+
+Online courses allow for some flexibility in how class time is integrated into your routine. Nevertheless, there is a
+well-structured and demanding workflow to which you are expected to adhere.
+
+This course is designed for a typical 12 week semester, but is being offered as an accelerated 8 week course. Students
+must be prepared for the pacing of the course.
+
+
+
+.. list-table:: Topic and Lab Schedule
+    :widths: 50 50 50 50
+    :header-rows: 1
+
+    * - Week Number
+      - Monday Date
+      - Topics
+      - Labs
+    * - Week 1
+      - June 24
+      - 1 -- 4
+      - 1 -- 2
+    * - Week 2
+      - July 1
+      - 5 -- 8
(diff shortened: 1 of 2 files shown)


twentylemon:
> asynch

bruh

site/outline/schedule.rst
```diff
@@ -0,0 +1,84 @@
+===============
+Course Schedule
+===============
+
+Online courses allow for some flexibility in how class time is integrated into your routine. Nevertheless, there is a
+well-structured and demanding workflow to which you are expected to adhere.
+
+This course is designed for a typical 12 week semester, but is being offered as an accelerated 8 week course. Students
```
twentylemon, line 8: 8 weeks? Why so long?

site/outline/schedule.rst
```diff
@@ -0,0 +1,84 @@
...
+
+
+.. list-table:: Topic and Lab Schedule
+    :widths: 50 50 50 50
+    :header-rows: 1
+
+    * - Week Number
+      - Monday Date
+      - Topics
+      - Labs
```
twentylemon, line 20: Are there already links on the page to these topics/labs? May be worthwhile to still link them here for that extra sexy USABILITY

twentylemon (approved):

site/outline/schedule.rst
jameshughes89: Since most of the entries on the page are ranges (e.g. 1 -- 4), it's tough to link nicely without spelling out each one. 

In other words, worse is better.

PR #634, opened 2024-08-26
:mortar_board: Assignmnet 1 stacbucks --- Clear up edge case on rectangle :microscope: 

jameshughes89:
### What

Clear up the edge case on when to count Starbucks locations --- only count them when they fall strictly between the lats/lons (only count when `lat_1 < starbucks_lat < lat_2`).

### Why

Created confusion in the past. 

### Testing

:-1: 

### Additional Notes

The reason for doing strictly between is because they align with the existing tests. Ultimately, however, it's arbitrary.


site/assignments/starbucks/starbucks.rst
```diff
@@ -136,12 +136,15 @@ Part 5 --- Counting Starbucks
...
 In order to calculate the density of Starbucks, the number of Starbucks within the specified area needs to be known.
-With the data available, the way to do this is to check each Starbucks' latitude & longitude and check if it falls
+With the data available, the way to do this is to check each Starbucks' latitude & longitude and see if it falls
 within the specified "rectangle". In other words, check if the Starbucks' latitude falls between the "rectangle's"
 latitudes and if the longitude falls between the "rectangle's" longitudes. See the below image for an example.
 
 .. image:: a1-LatLongSquareCount.png
 
+For the purpose of this assignment, only count Starbucks if they are strictly between the specified latitudes and
+longitudes; don't count them if they fall on any of the latitude/longitude lines.
```
twentylemon, line 146: I'm not sure if first years know the "strictly" jargon. May want to be verbose and say something like "that means < and not <="

twentylemon (approved):

site/assignments/starbucks/starbucks.rst
jameshughes89: I think by this time I'll have used it, but I'll add something to make it clearer.

PR #643, opened 2024-10-10
🎓 Assignment zombies --- Move due date back 1 week 🔬 

jameshughes89:
### What

Move the due date back 1 week

### Why

Im not sure why I picked 21st. Did I forget reading week was a thing? Defo not far enough along. I intended for the due date to be the Monday after the week after reading week. 

### Testing
Nah


twentylemon:
> Did I forget reading week was a thing?

It used to not be a thing, so. maybe.

twentylemon (approved):

PR #654, opened 2024-11-29
Update test_course.py to add tests for zero students

YourLocalMedic:
added tests for course.contains() and course.remove() where there are zero students in the course


twentylemon (changes requested):
Tests in this file are grouped by the list cardinality. Empty list tests are at the top of the file; where there is already a test for removing from the empty list.

Please move the `contains` test to be with the other empty list tests, and remove the duplicate test.

twentylemon (approved):
🚀

PR #669, opened 2025-04-16
:arrow_up: Spring 2025 --- Move dates of A4 and Exam  up

jameshughes89:
### What

Move dates of assignment 4 and the exam up. 

### Why

161 is to be offered in the Spring and 162 in the Summer. 

Although the week of the 23 is the exam week for Spring, the Summer semester starts that week too. Since 161 is prereq for 162, it would be better to have the exam as early as possible.


twentylemon (approved):
/shrug seems like it won't make a difference. Assignments being due on a sunday also feels like a 🖕

PR #677, opened 2025-08-25
Fall 2025 --- Course Setup

jameshughes89:
### What

Setup the course for the Fall 2025 offering

### How

- Update days/times
- Remove schedule from summer
- Add wording on lecture activities

### Testing

Built local and it didn't yell at me


site/outline/outline.rst
```diff
@@ -256,6 +251,42 @@ any questions they have regarding the material.
 
 
 
+Lecture & Lab Activities
+========================
+
+Lcture & lab activities will take during certain lectures and labs.
```
twentylemon, line 257: typo _Lecture_

May be useful to pull up this text blob from git history, or store them in python or something and have rst stitch the files together (iunno if that is possible).

jameshughes89: I actually pulled it from the history of another, but changed it slightly which caused the typo. 

Good catch.

PR #729, opened 2026-04-14
:mortar_board: Spring 2026 --- Setup course 

jameshughes89:
### What

Setup the course website for Spring 2026 asynchronous offering. 

### How

I looked at the diff from setting up Fall 2025. 

### Testing

Built local, looks good.

### DIFF
--- a/site/outline/schedule.rst
+++ b/site/outline/schedule.rst
@@ -1,0 +1,79 @@
+***************
+Course Schedule
+***************
+
+ Online courses allow for some flexibility in how class time is integrated into your routine. Nevertheless, there is a
+ well-structured and demanding workflow to which you are expected to adhere.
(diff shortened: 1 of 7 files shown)


site/outline/schedule.rst
```diff
@@ -0,0 +1,79 @@
+***************
```
twentylemon, line 1: Do you have the old one from the last time this went async? Or was that a different course

twentylemon (approved):

site/outline/schedule.rst
jameshughes89: Same course, but this is a modified copy/paste of the old one. This year they moved to make the courses 7 weeks instead of 8, so I had to smush things.

PR #743, opened 2026-04-16
:lipstick: Assignment FAQ --- Frame as questions and add AI question :robot: 

jameshughes89:
### What

Frame all FAQ points as questions
Add an AI question


twentylemon (approved):
Looks good — one minor grammar nit: "Why does some of the code in the provided functions **looks** like magic?" should be "**look**" (auxiliary verb "does" requires the bare infinitive). Worth fixing in a follow-up.

PR #745, opened 2026-04-20
:arrow_up: Python Version --- Set >= 3.12

jameshughes89:
### What

1. Change python 3.11 -> 3.12. 
2. Update black to use 312 (it was 310). 
3. Change the README to use `python3` instead of `python3.XX`

### Why

RE 1
I just installed Ubuntu 24 and it comes with 3.12, so I figured, _eh, why not_. 

One may ask _why not use 3.14_, but one package may have an issue with 4.14.1 (networkx), so, I will leave that for now. 

RE 2 
(a) it was set to 310 and should have been set to 310. (b) because now it matches 3.12. 

RE 3
Simplifies it for the repo. I know this _could_ create issues, but I think the simplification for the students is better. 

### Testing

was able to get everything installed and tests passed. 

### Additional Notes

I have **not** run `format` on the repo. This is because I have not been running it for a while and there are a lot of things that would be updated. I will do a separate PR for that.

### DIFF
--- a/.github/workflows/sphinx-build.yml
+++ b/.github/workflows/sphinx-build.yml
@@ -17,7 +17,7 @@ jobs:
     - name: Set up Python
       uses: actions/setup-python@v6
       with:
-        python-version: 3.11
+        python-version: 3.12
     - name: Install Dependencies
       run: |
         pip install -e .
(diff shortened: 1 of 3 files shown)


jameshughes89:
Draft until I try it on 3.14. Everything may end up being fine.

jameshughes89:
> Draft until I try it on 3.14. Everything may end up being fine.

Actually, I think 3.12 is the best option for now. Less potential issues.

twentylemon (approved):
Straightforward version bump. `black` target-version was stale at `py310`; now correctly `py312`. README simplification (`python3` vs `python3.11`) is good for students. CI passes.

PR #747, opened 2026-04-20
:mortar_board: Topic intro --- Warn about AI

jameshughes89:
### What

Explain that AI is prohibited. Also explain that struggling through the content is important. 

### Why

I took a laissez faire attitude about AI before, and that's not working. I am now more clearly saying no. I expect many will still use it, but maybe this will have an impact. 

### Testing

YOLO

### DIFF
--- a/site/topics/intro/intro.rst
+++ b/site/topics/intro/intro.rst
@@ -175,6 +175,7 @@ What about Generative AI?
 * When talking about Googling, I am not talking about generative AI
 
     * The friction of searching for a solution is an important part in learning
+    * Formulating the right question, reading various answers, piecing together an answer, is part of learning
 
 * AI is a great tool for programmers
 * AI is *not* a great tool for those learning to program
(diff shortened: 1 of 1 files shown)


twentylemon:
lemme get claude to review

twentylemon (approved):
Suggestion on the new "What about Generative AI?" section: it currently leans entirely on "struggle is necessary." Consider adding a line that names *why* the tools are fine for working programmers but corrosive for learners — something like:

> Spotting when AI output is wrong, shallow, or hallucinated requires the very expertise you're here to build — and you don't build that expertise by letting the tool do the reps.

This grants the tools' real value (which students already see their working-programmer friends getting) instead of only forbidding them, which I think lands harder. Drop in under the existing "AI is *not* a great tool for those learning to program" bullet if it resonates.

twentylemon:
to be clear, ^ is my opinion but claude's writing just 'cause

jameshughes89:
I'm for it.

PR #751, opened 2026-04-21
:lipstick: Site --- Run format and update verify

jameshughes89:
### Related Issues or PRs

Talked about in #745 

### What

1. Run format
2. Update flake8 runes to ignore E712 (using `==` bools vs `is`)

### Why

1. I have been forgetting to run this. 
2. A little conflicted on this as, perhaps I should be using `is` on bools. BUT I think this would add unnecessary mental load to the students at this stage, especially when they are using this as a simple testing trick. 

### Additional Notes

May need to re-run if/when #745 merges (though, i should be re-running all the time anyways).


twentylemon (approved):
Primarily mechanical formatting changes. The E712 ignore (`==` vs `is` for booleans) is a reasonable call for a teaching context — avoids introducing that distinction before students are ready. Note: if #745 merges before this, it may be worth re-running `format` since the black target-version changes. CI passes.

PR #752, opened 2026-04-21
:robot: Repo --- Add `sphinx-lint` to project

### LINKED ISSUE #7: ADD code formatting
Need to incorporate code formatting for the repo. Both the Java and ReStructuredText.

jameshughes89:
### Related Issues or PRs

Closes #7 

### What

Add `sphinx-lint` to do checks on the the rst files and add this to the project script verify. I also renamed the project scripts to not specifically mention code as the formatter would run on md (which I guess one _could_ argue is "code") and the verify now runs on the rst stuff (which is "code" also, I guess). 

I also ran the verify and fixed all the issues. All were variations of whitespace issues except for two python syntax issues from the assignments. 

### Why

Cleaner rst files now

### Testing

Rebuilt, everything is good. 

### Additional Notes

the linter will complain about the python syntax in the assignment files, but this is expected since they are incomplete assignment code for the students. 

At first I added the setting to ignore python syntax, but then it would ignore the syntax for everything, which is no good. I then ignored the assignments folders, but that is also not great as I still want to check everything else in there. In the end, I will simply accept and ignore the 2 warnings that it gives me.


twentylemon:
> the linter will complain about the python syntax in the assignment files, but this is expected since they are incomplete assignment code for the students.
> 
> At first I added the setting to ignore python syntax, but then it would ignore the syntax for everything, which is no good. I then ignored the assignments folders, but that is also not great as I still want to check everything else in there. In the end, I will simply accept and ignore the 2 warnings that it gives me.

There's usually ways to tell linters/formatters to ignore blocks, like a `#format: on ... #format: off` thing. Is there any such tool here?

Having a linter where the expected outcome is lint errors is kinda pointless.

jameshughes89:
> > the linter will complain about the python syntax in the assignment files, but this is expected since they are incomplete assignment code for the students.
> > At first I added the setting to ignore python syntax, but then it would ignore the syntax for everything, which is no good. I then ignored the assignments folders, but that is also not great as I still want to check everything else in there. In the end, I will simply accept and ignore the 2 warnings that it gives me.
> 
> There's usually ways to tell linters/formatters to ignore blocks, like a `#format: on ... #format: off` thing. Is there any such tool here?
> 
> Having a linter where the expected outcome is lint errors is kinda pointless.

Seems that there are no options for that. 

Two potential simple solutions:
1. Add `pass # Delete this line and write your code here` to each of the specific complaints 
3. Specifically ignore those files. 

Pro for 1 is makes the problem go away and is "correct", but con is (a) students may get confused, delete the comment,  but leave the `pass` 

Pro for 2 is makes the problem go away, but potential other issues in assignment files could get missed. 

I suppose I lean towards option 1, or, as it is, just sadly live with the 2 lint complaints.

twentylemon:
> but con is (a) students may get confused, delete the comment, but leave the `pass`

this feels like reaching

jameshughes89:
> > but con is (a) students may get confused, delete the comment, but leave the `pass`
> 
> this feels like reaching

I was thinking on it more, and I think I will have shown them pass before this time anyways, so they should know its to be removed.

twentylemon (approved):
Closes #7. Discussion resolved — the `pass # Delete me and write your code here` approach is the right call. Note: this PR and #753 both rename `run_code_formatters`/`run_code_verification` in `cli.py`; whichever merges second will need a rebase. Whitespace fixes are clean. CI passes.

PR #754, opened 2026-04-21
:lipstick: Getting Set --- Simplify wording on turning off AI :microscope: 

jameshughes89:
### What

Simplify the wording of why we are turning off the colab AI features. 

### Why

1. Cleaner
2. It mentioned that it is often incorrect. This isn't really true. It can be _incorrect_ in that it would solve things in "better" ways than what I want them to do for learning purposes. But its still correct. Either way, simpler words.


twentylemon (approved):
Small, clean improvement. The original wording implied AI tools are "incorrect" which wasn't quite right; the new wording ("tries to help with things you're here to learn") is more accurate and more constructive. CI passes.

PR #755, opened 2026-04-21
:memo: Topic Intro --- Note on hand writing notes :microscope: 

jameshughes89:
### What

Emphasize hand written notes. 

Also sneak in a CS -> CSCI 

### Why

Hardly anyone does it (I think the website actually causes this) but it will probably help with their learning.


twentylemon (approved):
Good pedagogical addition. The note on handwritten notes is well-motivated and appropriately gentle (recommends rather than mandates). `CS 161` → `CSCI 161` fix is correct. CI passes.

PR #756, opened 2026-04-22
:feather: Topic Intro --- Soften Language

jameshughes89:
### What

Soften the language a little throughout the topic 

### Why

I consistently get feedback that I am/my content is too intimidating. Perhaps this will help a little. 

### How

A little here and there. I don't think I change the message at all, just the words a bit to be framed more positively/less negative.


twentylemon (approved):
Measured, thoughtful changes throughout. The reframe from "So I Should Be Terrified?" → "Is This Course Hard?" removes unnecessary alarm without downplaying the course. Adding "this is the nature of programming, even for experienced programmers" is particularly good. CI passes.

PR #757, opened 2026-04-22
:lipstick: Topics --- For Next Class -> For Next Topic

jameshughes89:
### Related Issues or PRs

Kinda related to #756 

### What

"For Next Class" -> "For Next Topic" 

Also change the last topic's to be "For Next Course" and link to Python vs Java. 

### Why

I don't like the word "class", and its for the topic anyways, not a specific class. 

### Testing

Built local, looks good, link works. 

### Additional Notes

The last topic originally said "for next class" and it referenced a recursion topic in the textbook. I had actually covered recursion in the past and have thought about re-adding it, but I dont think there is enough time as is.


twentylemon (approved):
"For Next Topic" is clearer than "For Next Class" (avoids ambiguity with lecture vs. topic). The last topic becoming "For Next Course" + CSCI 162 link is a nice cap. Change is consistently applied across all 24 topic files. CI passes.

PR #758, opened 2026-04-22
:broom: Topic Values, Variables, Types --- Clean

jameshughes89:
### What

1. Remove early debugging section (too early and already talked about in intro)
2. Remove formal vs natural language section (not important here)
3. Remove terminology section (not important here)
4. General clean up and nicer words where possible

### Why

General maintenance. 

I could argue to keep the debugging, language, and terminology topics, but I think simplifying is more valuable here.


twentylemon (approved):
Removing the debugging, formal/natural language, and terminology sections simplifies the topic; the rationale (covered elsewhere or not important at this stage) is sound. Wording improvements are clean. CI passes.

PR #760, opened 2026-04-23
:broom: Run format and verify 

jameshughes89:
### What

Run format and verify with all the updates. 

### Why

consistency

### Additional Notes

The only thing I can see as a potential problem is, in general, my comments are `# this is some comment`, however, when I have a line that is commented out that is to be un-commented out by the students, I would have something like this `#someTestCode...`. 

The formatter would add a space, which is not a problem, but I can see some students only deleting the `#`, but leaving the space ` `. This could then cause a headache for them. 

For now I assume its not actually going to be an issue, but I am noting it here.


site/assignments/tic-tac-toe/asn3.py
```diff
@@ -24,15 +25,15 @@ def setup_game(size: int) -> list:
     """
 
 
+# assert [[" "]] == setup_game(1)
```
twentylemon, line 28: Does it make sense to have global flags to enable/disable these tests? You could do `if f(..) is not None` (or similar) for each. Then no comments needed at all

twentylemon (approved):

site/assignments/tic-tac-toe/asn3.py
jameshughes89: Although I like the idea in general, I think it has more potential to be annoying for the students at this stage.

PR #764, opened 2026-04-24
:broom: Topic functions --- Clean, simplify, soften 

jameshughes89:
### What

Clean up the functions topic 

### Testing

:+1: 

### Additional Notes

This one had more changes than the other cleans. 

The most noteworthy ones I think are:
1. remove general form of a function. It didnt make sense where I had it. By the time we get to it, it is probably obvious to them. I could have moved it earlier, but I think the topic is better without it. Less is more. 
2. My execution flow had 2 examples of the function being called before. I dropped it down to 1. I used to only have 1 adn added 2 a year or so ago. On one hand, there is some value in emphasizing how execution jumps around, but I also dont think its necessary. Less is more here.

### DIFF
--- a/site/topics/functions/functions.rst
+++ b/site/topics/functions/functions.rst
@@ -325,7 +325,7 @@ Creating Abstraction
     * Or, do you need to think about neurons firing and ion pumps to move your arm?
 
 * You are already experts at this in real life
-* Unfortunately, however, it is not simple to just start creating abstraction in your code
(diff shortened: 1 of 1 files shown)


twentylemon (approved):
`` `global` `` at line 456 — double backtick? it's more common in the file.

jameshughes89:
> `` `global` `` at line 456 — double backtick? it's more common in the file.

Good catch. I think it must be 2x `

PR #767, opened 2026-04-24
:broom: Topic if-else --- Clean and Soften

jameshughes89:
### What

Clean and soften 

### Why

General maintenance

### DIFF
--- a/site/topics/if-else/if-else.rst
+++ b/site/topics/if-else/if-else.rst
@@ -27,6 +27,9 @@ Conditional Expressions
 
 * The key now is to tell Python that, ``if`` some condition is ``True``, do something
 
+    * ``if some_condition:``
+    * The code to be run if the condition is ``True```` is
+
 .. code-block:: python
     :linenos:
     :emphasize-lines: 11
--- a/site/topics/if-else/if-else.rst
+++ b/site/topics/if-else/if-else.rst
@@ -331,7 +331,7 @@ Exclusive Alternatives
         * ``True``
 
     * Since the expression is evaluated to ``True``, the indented code is run
-    * Updates ``letter_grade`` the value ``"A"``
+    * *Updates* ``letter_grade`` the value ``"A"``
     * ...
     * This continues for each ``if``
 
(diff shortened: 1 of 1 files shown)


site/topics/if-else/if-else.rst
```diff
@@ -420,6 +427,43 @@ Nesting Conditionals
...
+            :linenos:
+
+
+            def is_eligible(a_number):
+                return a_number % 3 == 0 and a_number % 5 == 0
+
+            if is_eligible(a_number):
+                # Some Code
+
+
```
jameshughes89, line 440: Overkill?

twentylemon (approved):
A couple little things:

- line 33: ` ``True```` ` has an extra pair of backticks.
- line 334: "Updates `letter_grade` the value `"A"`" reads a bit off — and line 325 still says "Assign `letter_grade` the value …", so the trace is inconsistent with itself. Maybe just keep "Assigns".
- re: "Overkill?" — the bullets are fine, but the two code blocks feel like a lot. By this point students know how to assign a variable and define a function, so the snippets are showing them what they already know. The bullet headings alone get the point across.

jameshughes89:
> A couple little things:
> 
> * line 33: `` ``True```` `` has an extra pair of backticks.

:+1: 

> * line 334: "Updates `letter_grade` the value `"A"`" reads a bit off — and line 325 still says "Assign `letter_grade` the value …", so the trace is inconsistent with itself. Maybe just keep "Assigns".

The update is intentional. Its showing a problem with the flow --- like 325 assigns the value A+, but because the next `if` triggers, it *updates* the value. I made a change here, but it was to emphasize the *Updates*. 

> * re: "Overkill?" — the bullets are fine, but the two code blocks feel like a lot. By this point students know how to assign a variable and define a function, so the snippets are showing them what they already know. The bullet headings alone get the point across.

:+1: It will accompany me in lecture anyways, so if it seems like they don't get it, I'll just show them. 

In general, at this stage, with the content having matured over the past few years (for lack of a better term), I have nothing against adding additional content if it helps, but I find myself wanting to trim more than anything.

PR #780, opened 2026-05-04
:broom: Topic Objects Methods --- Clean and Soften

jameshughes89:
### What

Clean and Soften. Also fixed up related src files from related topics.


twentylemon (approved):
Sub-bullet indentation is inconsistent with the rest of the file (6 spaces vs 4).

